PR #65, opened 2019-06-18
starting controller

HagarMeir:


internal/bft/controller.go
```diff
@@ -0,0 +1,283 @@
...
+		<-c.viewAbortChan
+	}
+	atomic.StoreUint64(&c.currViewNumber, newViewNumber)
+	c.controllerStart = false
+	c.stopWG.Add(1)
+	go func() {
+		defer c.stopWG.Done()
+		c.startView()
+	}()
+	<-c.viewReadyChan
```
yacovm, line 192: are you sure you need the **viewReadyChan** ? 
Seems like iAmTheLeader is only effected by that atomic unit64 replacement done in line 185, so is it possible to remove this field?

internal/bft/controller.go
```diff
@@ -0,0 +1,283 @@
...
+		go func() {
+			defer c.stopWG.Done()
+			c.leader()
+		}()
+	}
+}
+
+func (c *Controller) propose() {
+	var validRequests [][]byte
+	for len(validRequests) == 0 { // no valid requests in this batch
```
yacovm, line 205: Nice corner case handling to avoid an empty batch!

It seems like [this bunch of code](https://github.com/SmartBFT-Go/consensus/blob/d64ac3d51ba2d27ee9628f8b95497f33cd5be96e/internal/bft/controller.go#L204-L223) is self contained. Can we please extract it to a function so the function will be smaller?

internal/bft/controller.go
```diff
@@ -0,0 +1,283 @@
...
+		for _, req := range requests {
+			err := c.Verifier.VerifyRequest(req)
+			if err != nil {
+				c.Logger.Warnf("Ignoring bad request: %v, verifier error is: %v", req, err)
+				continue
+			}
+			validRequests = append(validRequests, req)
+		}
+	}
+	var metadata []byte // TODO get metadata
```
yacovm, line 224: You can do the following:

```
	// This should be a method of the view.
	// we should not lock externally
	c.currView.lock.RLock()
	propSeq := c.currView.ProposalSequence
	c.currView.lock.RUnlock()
	md := &protos.BlockMetadata{
		ViewId: c.currViewNumber,
		LatestSequence: propSeq,
	}
	metadata, err := proto.Marshal(md)
	if err != nil {
		c.Logger.Panicf("Faild marshaling metadata: %v")
	}
```

internal/bft/controller.go
```diff
@@ -0,0 +1,283 @@
...
+	end := c.currView.Start()
+	c.viewLock.Unlock()
+	c.viewReadyChan <- struct{}{}
+	end.Wait()
+	c.viewAbortChan <- struct{}{}
+}
+
+// ViewChanged makes the controller abort the current view and start a new one with the given number
+func (c *Controller) ViewChanged(newViewNumber uint64) {
+	if !c.controllerStart {
```
yacovm, line 178: this is used as a "one time operation", right?

I think it's better to decompose this to 2 functions:
- One that aborts the previous view
- One that starts the next view

And call only the latter in **Start()**.

internal/bft/controller.go
```diff
@@ -0,0 +1,283 @@
...
+		PrevHeader:       []byte{0}, // TODO start with real prev header
+	}
+
+	c.viewLock.Lock()
+	c.currView = view
+	c.Logger.Debugf("Starting view with number %d", atomic.LoadUint64(&c.currViewNumber))
+	end := c.currView.Start()
+	c.viewLock.Unlock()
+	c.viewReadyChan <- struct{}{}
+	end.Wait()
```
yacovm, line 172: Is it possible to use this **Future** to simplify the whole view lifecycle orchestration?
Perhaps you can have this **Future** a field of the controller, and use it to wait for the view termination when you invoke **ViewChanged** and hopefully you can then get rid of some of the channels of **viewReadyChan** and **viewAbortChan**?

internal/bft/controller_test.go
```diff
@@ -0,0 +1,220 @@
...
+			end := controller.Start(1)
+			<-verifyLog
+			controller.Stop()
+			end.Wait()
+		})
+	}
+
+}
+
+func TestControllerViewChanged(t *testing.T) {
```
yacovm, line 86: where are the assertions?

internal/bft/controller_test.go
```diff
@@ -0,0 +1,220 @@
...
+	end := controller.Start(1)
+
+	controller.ViewChanged(2)
+	controller.ViewChanged(3)
+
+	controller.Stop()
+	end.Wait()
+}
+
+func TestControllerLeader(t *testing.T) {
```
yacovm, line 104: ditto

internal/bft/controller_test.go
```diff
@@ -0,0 +1,220 @@
...
+		N:       4,
+		Logger:  log,
+		Batcher: batcher,
+	}
+	end := controller.Start(1)
+	controller.Stop()
+	end.Wait()
+}
+
+func TestLeaderPropose(t *testing.T) {
```
yacovm, line 121: same. This is a nice scenario, and I believe it works, but we also need assertions at some point. 
For example - you can check that AssembleProposal was called with such and such requests, and you can check that Broadcast() was called with messages of the right view and sequence.

yacovm (approved):

yacovm:
Nice work! :+1:  

See comments to address later on. Merging in the meanwhile.

internal/bft/controller.go
HagarMeir: I don't want the leader to start proposing before the view is started, before the `c.currView` has the right value, and so I added the `viewReadyChan`.
Does that makes sense?

internal/bft/controller.go
HagarMeir: Sure

internal/bft/controller.go
HagarMeir: True
Will do

internal/bft/controller.go
HagarMeir: OK

internal/bft/controller.go
HagarMeir: Will try
Although it does make sense to work with the channels in most cases

internal/bft/controller.go
yacovm: yes it does.

PR #77, opened 2019-06-24
adding a simple request pool

HagarMeir:


internal/bft/controller.go
```diff
@@ -120,15 +115,18 @@ func (c *Controller) leaderID() uint64 {
...
 	c.Logger.Debugf("The number of nodes (N) is %d, F is %d, and the quorum size is %d", c.N, f, q)
 	return q
 }
 
 // SubmitRequest submits a request to go through consensus
 func (c *Controller) SubmitRequest(request []byte) {
-	c.RequestPool.Submit(request)
+	err := c.RequestPool.Submit(request)
+	if err != nil {
+		c.Logger.Warnf("Request %v was not submitted, error: %v", request, err)
```
C0rWin: Not sure how is it helpful to get slices of bytes printed. Probably need to print compact and human consumable representation of it, wdyt?

HagarMeir: should I invoke requestInspector here to print out the requestInfo?

internal/bft/requestpool.go
```diff
@@ -0,0 +1,97 @@
...
+			err := fmt.Sprintf("a request with ID %v and client ID %v already exists in the pool", reqInfo.ID, reqInfo.ClientID)
+			rp.Log.Errorf(err)
+			return fmt.Errorf(err)
+		}
+	}
+	rp.queue = append(rp.queue, req)
+	return nil
+}
+
+// NextRequests return the next requests to be batched
```
C0rWin, line 69: We are not going to consent on next batch until previous was decided, right?

internal/bft/requestpool.go
```diff
@@ -0,0 +1,97 @@
...
+		ID:       reqInfo.ID,
+		Request:  request,
+		ClientID: reqInfo.ClientID,
+	}
+	if err := rp.semaphore.Acquire(context.Background(), 1); err != nil {
+		return fmt.Errorf("error in acquiring semaphore: %v", err)
+	}
+	rp.lock.Lock()
+	defer rp.lock.Unlock()
+	for _, existingReq := range rp.queue {
```
C0rWin: this might be a nit, but could we probably introduce a map[string]bool to optimize this check? instead of iterating checking whenever key is in the map.

internal/bft/requestpool.go
```diff
@@ -0,0 +1,97 @@
...
+	rp.lock.Lock()
+	defer rp.lock.Unlock()
+	removed := false
+	for i, existingReq := range rp.queue {
+		if existingReq.ClientID == request.ClientID && existingReq.ID == request.ID {
+			rp.Log.Infof("Removed request %v from request pool", request)
+			rp.queue = append(rp.queue[:i], rp.queue[i+1:]...)
+			removed = true
+			rp.semaphore.Release(1)
+			break
```
C0rWin: why not return?

internal/bft/requestpool.go
```diff
@@ -0,0 +1,97 @@
...
+		return rp.queue
+	}
+	return rp.queue[:n]
+}
+
+// RemoveRequest removes the given request from the pool
+func (rp *RequestPool) RemoveRequest(request Request) error {
+	rp.lock.Lock()
+	defer rp.lock.Unlock()
+	removed := false
```
C0rWin: if you return at line 88, then no need for `removed` variable.

C0rWin (changes requested):
I've left a couple of comments, I think they need to be addressed/answered :)

internal/bft/requestpool.go
HagarMeir: True

internal/bft/controller.go
C0rWin: I think it will make more sense, cause overwise I do not see much value in printing this information.

internal/bft/requestpool.go
C0rWin: actually it could be a map[string]*Request

internal/bft/requestpool.go
```diff
@@ -0,0 +1,99 @@
...
+		return rp.queue
+	}
+	return rp.queue[:n]
+}
+
+// RemoveRequest removes the given request from the pool
+func (rp *RequestPool) RemoveRequest(request Request) error {
+	rp.lock.Lock()
+	defer rp.lock.Unlock()
+	existStr := fmt.Sprintf("%v~%v",request.ClientID, request.ID)
```
C0rWin: I think it will be better if you can do instead:
```

if _, exist := rp.existMap[existStr]; !exist {
       err := fmt.Sprintf("Request %v is not in the pool at remove time", request)
	rp.Log.Warnf(err)
	return fmt.Errorf(err)
}

for i, existingReq := range rp.queue {
     if !existingReq.ClientID == request.ClientID || !existingReq.ID == request.ID {
          continue
     }
     rp.Log.Infof("Removed request %v from request pool", request)
     rp.queue = append(rp.queue[:i], rp.queue[i+1:]...)
     delete(rp.existMap, existStr)
     rp.semaphore.Release(1)
     return nil
}
```

C0rWin:
Seemed it failed cause of formatting:

> The following files needs gofmt:
./internal/bft/requestpool.go
./internal/bft/controller.go
make: *** [all] Error 1

PR #78, opened 2019-06-25
add batcher

HagarMeir:


internal/bft/batcher.go
```diff
@@ -0,0 +1,55 @@
...
+				return b.buildBatch(remainderOccupied, currBatch)
+			}
+			time.Sleep(b.Timeout / 100)
+		}
+	}
+}
+
+func (b *Bundler) buildBatch(remainderOccupied int, currBatch [][]byte) [][]byte {
+	reqs := b.Pool.NextRequests(b.BatchSize - remainderOccupied)
+	reqsBytes := make([][]byte, 0)
```
C0rWin: why do you need it? you can append to currBatch directly, no? also it will help to add docs here explaining side effects of this function -  mutation of current batch for example. we are not exposing it yes, but still to make it more clear.

internal/bft/controller.go
```diff
@@ -278,10 +286,16 @@ func (c *Controller) Stop() {
 }
 
 // Decide delivers the decision to the application
-func (c *Controller) Decide(proposal types.Proposal, signatures []types.Signature) {
+func (c *Controller) Decide(proposal types.Proposal, signatures []types.Signature, requests []types.RequestInfo) {
 	// TODO write to WAL?
-	// TODO remove and stop timeouts of included requests?
 	c.Application.Deliver(proposal, signatures)
+	c.Logger.Infof("Node %d delivered proposal", c.ID)
```
C0rWin: I think this is frequent and normal mode operation, no need to log this in Info level, probably Debug?

PR #100, opened 2019-07-10
Fix bft.Comm dependency API

C0rWin:
Change SendTransaction API of bft package to receive slice of bytes and
not proto message.

### DIFF
--- a/examples/naive_chain/node.go
+++ b/examples/naive_chain/node.go
@@ -102,8 +102,8 @@ func (n *Node) SendConsensus(targetID uint64, message *protos.Message) {
 	n.out[int(targetID)] <- message
 }
 
-func (n *Node) SendTransaction(targetID uint64, message *protos.Message) {
-	n.out[int(targetID)] <- message
+func (n *Node) SendTransaction(targetID uint64, request []byte) {
+	// TODO: Handle send transaction request
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/dependencies.go
```diff
@@ -17,7 +17,7 @@ type Application interface {
 type Comm interface {
 	BroadcastConsensus(m *protos.Message) // broadcast message to others (not including yourself)
 	SendConsensus(targetID uint64, m *protos.Message)
-	SendTransaction(targetID uint64, m *protos.Message)
+	SendTransaction(targetID uint64, request []byte)
```
C0rWin: test

examples/naive_chain/node.go
```diff
@@ -102,8 +102,8 @@ func (n *Node) SendConsensus(targetID uint64, message *protos.Message) {
 	n.out[int(targetID)] <- message
 }
 
-func (n *Node) SendTransaction(targetID uint64, message *protos.Message) {
-	n.out[int(targetID)] <- message
+func (n *Node) SendTransaction(targetID uint64, request []byte) {
+	// TODO: Handle send transaction request
```
C0rWin: test

C0rWin: sdf

pkg/api/dependencies.go
C0rWin: sss

PR #115, opened 2019-07-14
Rename request pool Start into NewPool

C0rWin:
Currently request pool introduces a function Start which basically
initializes the structure. This commit renames it to better reflect its
purpose (adding a few godocs).


internal/bft/controller.go
```diff
@@ -265,7 +265,7 @@ func (c *Controller) leader() {
 	}
 }
 
-// Start the controller
+// NewPool the controller
```
tock-ibm: this refactoring is not in place

C0rWin: thanks, fixed

PR #128, opened 2019-07-17
Generate the mocks

tock-ibm:
Generate the mocks cleanly instead of a double definition
of api interfaces.

Signed-off-by: Yoav Tock <tock@il.ibm.com>


internal/bft/mocks/generate_api_mocks.go
```diff
@@ -0,0 +1,20 @@
+// Copyright IBM Corp. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package mocks
+
+// Generate mocks for a collection of interfaces that are defined in api/dependencies.go
+
+//go:generate mockery -dir $GOPATH/src/github.com/SmartBFT-Go/consensus/pkg/api -name Verifier -case underscore -output .
```
C0rWin: I do not think that referencing an absolute path is a good way to go.

internal/bft/requestpool.go
```diff
@@ -15,13 +15,17 @@ import (
...
 )
 
-type requestTimeoutHandler interface {
+//go:generate mockery -dir . -name RequestTimeoutHandler -case underscore -output ./mocks/
 
-	// called when a request timeout expires
-	onRequestTimeout(request []byte)
+// RequestTimeoutHandler defines the methods called by request timeout timers created by time.AfterFunc.
+// This interface is implemented by the bft.Controller.
+type RequestTimeoutHandler interface {
```
C0rWin: I guess this change doesn't related to the moving mocks generation into separate central place, right?

PR #177, opened 2019-07-25
some cleanup

HagarMeir:

### DIFF
--- a/internal/bft/requestpool.go
+++ b/internal/bft/requestpool.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	DefaultRequestTimeoutMillis = 10000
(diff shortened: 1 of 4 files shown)


internal/bft/requestpool.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	DefaultRequestTimeoutMillis = 10000
+	DefaultRequestTimeoutMillis = 60000
```
tock-ibm, line 22: There is an Option object you give to the Pool on construction, where you can adjust timeouts for the tests etc, here it is meant to be a "reasonable" value for production.

HagarMeir: we should close the pool when testing
and then we can revert back to 10000

Issue #152, opened 2019-07-21
Leader ID


HagarMeir:
Currently the current leader ID is selected in the following manner:
`c.currViewNumber % c.N`

However, for fairness, this assumes that the ids of the nodes are sequential. 
If it is not the case then we should change the leader ID selection to something a little more sophisticated, maybe do something similar to what is implemented in BFT SMaRt.

PR #191, opened 2019-07-28
Generalize leader election

C0rWin:
This commit removes assumption of continious relicas ids

### DIFF
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -7,6 +7,7 @@ package bft
 
 import (
 	"math"
+	"sort"
 	"sync"
 	"time"
 
@@ -108,7 +109,11 @@ func (c *Controller) iAmTheLeader() (bool, uint64) {
 // thread safe
 func (c *Controller) leaderID() uint64 {
 	// TODO use ids order (similar to BFT Smart)
-	return c.getCurrentViewNumber() % c.N
+	nodes := c.Comm.Nodes()
+	sort.Slice(nodes, func(i, j int) bool {
+		return nodes[i] < nodes[j]
+	})
+	return nodes[c.getCurrentViewNumber()%c.N]
 }
 
 // computeQuorum calculates the quorums size Q, given a cluster size N.
--- a/examples/naive_chain/node.go
+++ b/examples/naive_chain/node.go
@@ -19,9 +19,6 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
-//type Ingress map[int]<-chan *smartbftprotos.Message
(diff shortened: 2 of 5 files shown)


internal/bft/controller.go
```diff
@@ -108,7 +109,11 @@ func (c *Controller) iAmTheLeader() (bool, uint64) {
 // thread safe
 func (c *Controller) leaderID() uint64 {
 	// TODO use ids order (similar to BFT Smart)
-	return c.getCurrentViewNumber() % c.N
+	nodes := c.Comm.Nodes()
```
yacovm: why it is important that they are monotonously ascending/descending? I don't think we really use any kind of assumption on their order anywhere...

internal/bft/controller.go
```diff
@@ -108,7 +109,11 @@ func (c *Controller) iAmTheLeader() (bool, uint64) {
...
+	sort.Slice(nodes, func(i, j int) bool {
```
yacovm, line 113: why it is important that they are monotonously ascending/descending? I don't think we really use any kind of assumption on their order anywhere...

internal/bft/controller.go
```diff
@@ -108,7 +109,11 @@ func (c *Controller) iAmTheLeader() (bool, uint64) {
 // thread safe
 func (c *Controller) leaderID() uint64 {
 	// TODO use ids order (similar to BFT Smart)
```
yacovm, line 111: the TODO can be removed now, no?

internal/bft/controller.go
C0rWin: it's important to have it deterministic

internal/bft/controller.go
C0rWin: yep

HagarMeir:
from BFT SMaRt:

```
/**
     * Deterministically elects a new leader, based current leader and membership
     * 
     * @return The new leader
     */
    public int getNewLeader() {

        int[] proc = SVController.getCurrentViewProcesses();
        int minProc = proc[0];
        int maxProc = proc[0];
                    
        for (int p : proc) {
            if (p < minProc) minProc = p;
            if (p > maxProc) maxProc = p;
        }
 
        
        do {
            currentLeader++;
            if (currentLeader > maxProc) {

                currentLeader = minProc;    
            }
        } while(!SVController.isCurrentViewMember(currentLeader));
        
        return currentLeader;
    }
```

HagarMeir:
Close this https://github.com/SmartBFT-Go/consensus/issues/152 when finished

PR #202, opened 2019-07-30
PersistedState initialized with WAL content

yacovm:

### DIFF
--- a/pkg/consensus/consensus.go
+++ b/pkg/consensus/consensus.go
@@ -27,16 +27,16 @@ type Consensus struct {
 	BatchSize    int
 	BatchTimeout time.Duration
 	bft.Comm
-	Application      bft.Application
-	Assembler        bft.Assembler
-	WAL              bft.WriteAheadLog
-	WALContent       [][]byte
-	Signer           bft.Signer
-	Verifier         bft.Verifier
-	RequestInspector bft.RequestInspector
-	Synchronizer     bft.Synchronizer
-	Logger           bft.Logger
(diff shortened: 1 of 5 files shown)


pkg/consensus/consensus.go
```diff
@@ -120,8 +121,9 @@ func (c *Consensus) BroadcastConsensus(m *protos.Message) {
 
 func (c *Consensus) NewProposer(leader, proposalSequence, viewNum uint64, quorumSize int) algorithm.Proposer {
 	persistedState := &algorithm.PersistedState{
-		Logger: c.Logger,
-		WAL:    c.WAL,
+		Entries: c.WALContent,
```
tock-ibm: This is only given/used once, are you sure you want to attach it to the PersistentState object forever?

pkg/consensus/consensus.go
```diff
@@ -30,6 +30,7 @@ type Consensus struct {
 	Application      bft.Application
 	Assembler        bft.Assembler
 	WAL              bft.WriteAheadLog
+	WALContent       [][]byte
```
tock-ibm: WALInitialContent maybe?

pkg/consensus/consensus.go
yacovm: Yes, because we're only going to append stuff into the state. I think it's good enough for now. I want to keep the logic of the read and write in the same file because it's easier to understand.

pkg/consensus/consensus.go
yacovm: done

Issue #109, opened 2019-07-11
Add periodical heart beats from leaders to followers


HagarMeir:

tock-ibm:
In the absence of requests, the leader may fail without anyone noticing until the next requests comes along, and triggers a timeout.

Create a mechanisms in which in the absence of requests the leader broadcasts periodic heartbeats. Followers maintain a heartbeat timeout timer, that is extended when a heartbeat is received. When a heartbeat expired the followers complain and start a view change process.

When requests are submitted, heartbeats should not be sent, and timeouts extended when requests are received (Backlog).

PR #240, opened 2019-08-06
Heartbeat monitor: follower role

tock-ibm:
Signed-off-by: Yoav Tock <tock@il.ibm.com>


internal/bft/heartbeatmonitor.go
```diff
@@ -90,35 +145,52 @@ func (hm *HeartbeatMonitor) StartLeader(view uint64, leaderID uint64) {
...
+	follower = hm.follower
+	if follower {
+		return view, follower, nil, false
+	}
+
+	heartbeat = &smartbftprotos.Message{
+		Content: &smartbftprotos.Message_HeartBeat{
+			HeartBeat: &smartbftprotos.HeartBeat{
+				View: view,
+				Seq:  0,
```
yacovm: why do we send sequence 0?  if we don't need it then we can just remove it no?

internal/bft/heartbeatmonitor.go
```diff
@@ -90,35 +145,52 @@ func (hm *HeartbeatMonitor) StartLeader(view uint64, leaderID uint64) {
...
+
+	if view == hm.view && follower == hm.follower {
+		hm.timer.Reset(hm.hbInterval)
+	}
 }
 
-// ProcessMsg handles an incoming heartbeat.
-func (hm *HeartbeatMonitor) ProcessMsg(sender uint64, msg *smartbftprotos.HeartBeat) {
-	// TODO
+func (hm *HeartbeatMonitor) prepareSend() (view uint64, follower bool, heartbeat *smartbftprotos.Message, doSend bool) {
```
yacovm, line 165: can't we just inline **prepareSend** into sendHeartbeat and then we don't need to return 4 arguments?

internal/bft/heartbeatmonitor_test.go
```diff
@@ -70,13 +70,198 @@ func TestHeartbeatMonitor_StartLeader(t *testing.T) {
...
 
 	hm.StartLeader(10, 12)
 	toWG1.Wait()
-	hm.Close()
 	hm.StartLeader(20, 12)
 	toWG2.Wait()
 	hm.Close()
 }
+
+func TestHeartbeatMonitor_Follower(t *testing.T) {
```
yacovm, line 86: Is it possible to somehow combine several sub-tests into a table test?

internal/bft/heartbeatmonitor_test.go
```diff
@@ -70,13 +70,198 @@ func TestHeartbeatMonitor_StartLeader(t *testing.T) {
...
+		hbTO := 500 * time.Millisecond
+		hm := bft.NewHeartbeatMonitor(log, hbTO, comm, handler)
+
+		handler.On("OnHeartbeatTimeout", mock.Anything, mock.Anything).Run(func(args mock.Arguments) {
+			t.Fail()
+		}).Return()
+
+		hm.StartFollower(10, 12)
+		hm.Close()
+		time.Sleep(hbTO * 2)
```
yacovm, line 149: why do we sleep when the test case ends?

internal/bft/heartbeatmonitor.go
```diff
@@ -25,16 +25,17 @@ type HeartbeatTimeoutHandler interface {
...
-	follower   bool
+	logger          api.Logger
+	hbTimeout       time.Duration
+	hbInterval      time.Duration
+	comm            Comm
+	handler         HeartbeatTimeoutHandler
+	mutex           sync.Mutex
+	timer           *time.Timer
+	view            uint64
+	leaderID        uint64
```
C0rWin, line 36: do you really need this one? View already has exposed LeaderID, why cann't you just use it instead?

internal/bft/heartbeatmonitor.go
```diff
@@ -57,25 +58,79 @@ func NewHeartbeatMonitor(
 	return hm
 }
 
-func (hm *HeartbeatMonitor) SetTimeoutHandler(handler HeartbeatTimeoutHandler) {
+// StartFollower will start following the heartbeats of the leader of the view.
+func (hm *HeartbeatMonitor) StartFollower(view uint64, leaderID uint64) {
+	hm.logger.Debugf("Starting heartbeat timeout on follower, duration: %s; leader: %d, view: %d",
+		hm.hbTimeout.String(), leaderID, view)
```
C0rWin, line 64: here for instance you can just call `view.LeaderID`

internal/bft/heartbeatmonitor.go
```diff
@@ -57,25 +58,79 @@ func NewHeartbeatMonitor(
...
-	//TODO
+	if hm.timer != nil {
+		hm.timer.Stop()
+	}
+	hm.view = view
+	hm.leaderID = leaderID
+	hm.follower = true
+	hm.lastInHeartbeat = time.Now()
+
+	hm.timer = time.AfterFunc(hm.hbTimeout, func() { hm.onHeartbeatTimeout(view, leaderID) })
```
C0rWin, line 77: `hm.onHeartbeatTimeout(view, leaderID)`

Is it timeout or hear beat tick?

internal/bft/heartbeatmonitor_test.go
tock-ibm: no

internal/bft/heartbeatmonitor.go
yacovm: The view doesn't really expose it, it is only initialized with it. 
We can compute the leader ID from N and from view, though.

tock-ibm: not really, the index of leaderID is exposed by the view number.

internal/bft/heartbeatmonitor.go
tock-ibm: That is true. I thought about about numbering the heartbeats but decided against it.

internal/bft/heartbeatmonitor.go
tock-ibm: yes we can, but i don't see any benefit... apart from making it less readable

internal/bft/heartbeatmonitor_test.go
tock-ibm: "close prevents timeout", to make sure close cancels the timeout

internal/bft/heartbeatmonitor.go
tock-ibm: As I said, the relation between leaderID and view is something that is computed above this unit; i don't think we should embed that logic here.

internal/bft/heartbeatmonitor.go
tock-ibm: heartbeat timeout, tick is called "heartbeat interval"

internal/bft/heartbeatmonitor.go
```diff
@@ -25,16 +25,17 @@ type HeartbeatTimeoutHandler interface {
...
-	view       uint64
-	leaderID   uint64
-	follower   bool
+	logger          api.Logger
+	hbTimeout       time.Duration
+	hbInterval      time.Duration
+	comm            Comm
+	handler         HeartbeatTimeoutHandler
+	mutex           sync.Mutex
+	timer           *time.Timer
```
yacovm, line 34: I am wondering if it is possible to have a time.Ticker instead of a time.Timer, and just take its time.Time channel and just use that to know when time passed. 

This way, you don't need to sleep in tests at all, and you can just inject an artificial time.Time channel into the struct, and just feed it with ticks to advance the time in the test.

tock-ibm: That is a possibility, but I am hesitant to do it just for the sake of a few seconds of testing. I'll need a thread to read the ticker, and so on... seems an overkill. I think that simplicity has its merits. However, if we decide to do this, I'll do it for the RequestPool timeouts as well in one go.

yacovm: I think the benefits outweigh the cost, because then you have more finer control on time timing of things, namely - now we can setup the view changer to notify you on a view change and ensure you picked up that information prior to you calling Complain on the view changer.

tock-ibm: I see the main advantage in tests. Lets put this into an issue for the backlog and revisit later.

tock-ibm:
See #109

PR #243, opened 2019-08-07
more to view change

HagarMeir:


internal/bft/viewchanger_test.go
```diff
@@ -108,15 +111,19 @@ func TestViewChangeProcess(t *testing.T) {
 	basicLog, err := zap.NewDevelopment()
 	assert.NoError(t, err)
 	log := basicLog.Sugar()
+	reqTimer := &mocks.RequestsTimer{}
+	reqTimer.On("StopTimers")
+	reqTimer.On("RestartTimers")
```
C0rWin, line 114: do you want to probably assert these functions are really being used and called?

PR #265, opened 2019-08-14
insert data to msg in view changer

HagarMeir:


pkg/types/types.go
```diff
@@ -19,7 +19,7 @@ type Proposal struct {
 	Payload              []byte
 	Header               []byte
 	Metadata             []byte
-	VerificationSequence int64 // TODO change to uint64
+	VerificationSequence int64 // int64 for ans1 marshaling
```
yacovm, line 22: asn1, not ans1

PR #268, opened 2019-08-15
handle last decision in view data

HagarMeir:

### DIFF
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -373,6 +373,7 @@ func (v *ViewChanger) processNewViewMsg(msg *protos.NewView) {
 	signed := msg.GetSignedViewData()
 	nodesMap := make(map[uint64]struct{}, v.N)
 	valid := 0
+	var maxLastDecisionSequence uint64
 	for _, svd := range signed {
 		if _, exist := nodesMap[svd.Signer]; exist {
 			continue // seen data from this node already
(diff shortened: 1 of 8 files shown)


internal/bft/viewchanger.go
```diff
@@ -306,6 +307,45 @@ func (v *ViewChanger) validateViewDataMsg(vd *protos.SignedViewData, sender uint
 		v.Logger.Warnf("%d got viewData message %v from %d, but %d is not the next leader", v.SelfID, rvd, sender, v.SelfID)
 		return false
 	}
+	return v.validateLastDecision(rvd, sender)
+}
+
+func (v *ViewChanger) validateLastDecision(vd *protos.ViewData, sender uint64) bool {
+	if vd.LastDecision == nil {
```
yacovm: if you don't have the last decision - shouldn't you panic? I thought we have to initialize you with the last decision in Consensus, no?

internal/bft/viewchanger.go
```diff
@@ -306,6 +307,45 @@ func (v *ViewChanger) validateViewDataMsg(vd *protos.SignedViewData, sender uint
...
+		v.Logger.Warnf("%d got viewData message %v from %d, but the last decision is not set", v.SelfID, vd, sender)
+		return false
+	}
+	if vd.LastDecision.Metadata == nil { // this is a genesis block, no signatures
+		v.Logger.Debugf("%d got viewData message %v from %d, last decision metadata is nil", v.SelfID, vd, sender)
+		// TODO handle genesis block
+		return true
+	}
+	if vd.LastDecisionSignatures == nil {
```
yacovm, line 340: so, len(vd.LastDecisionSignatures) returns 0 in case it's nil, no?

internal/bft/viewchanger.go
```diff
@@ -306,6 +307,45 @@ func (v *ViewChanger) validateViewDataMsg(vd *protos.SignedViewData, sender uint
...
+		v.Logger.Warnf("%d got viewData message %v from %d, but the last decision signatures is not set", v.SelfID, vd, sender)
+		return false
+	}
+	numSigs := len(vd.LastDecisionSignatures)
+	if numSigs < v.quorum {
+		v.Logger.Warnf("%d got viewData message %v from %d, but there are only %d last decision signatures", v.SelfID, vd, sender, numSigs)
+		return false
+	}
+	for _, sig := range vd.LastDecisionSignatures {
```
yacovm, line 349: you need to de-duplicate the signatures, otherwise I can take a single signed message and send it many times, no?

PR #282, opened 2019-08-22
wire scheduler into library

yacovm:


pkg/consensus/consensus.go
```diff
@@ -83,6 +83,7 @@ func (c *Consensus) Start() {
 	cpt := types.Checkpoint{}
 	cpt.Set(c.LastProposal, c.LastSignatures)
 
+	viewChangerClock := make(chan time.Time)
```
HagarMeir, line 86: how will the `viewChangerClock` tick?

PR #285, opened 2019-08-22
stop signing in prepares

yacovm:


internal/bft/view.go
```diff
@@ -452,13 +411,10 @@ func (v *View) processPrepares() Phase {
 		},
 	}
 
-	signaturesByIDs[v.SelfID] = ourPrepareSignature
-
 	preparedProof := &protos.SavedMessage{
```
HagarMeir, line 414: just save the commit now instead of preparedProof

smartbftprotos/messages.proto
```diff
@@ -38,15 +38,6 @@ message Prepare {
 message PreparedProof {
```
HagarMeir: Don't think we need this

HagarMeir (approved):

PR #286, opened 2019-08-22
Sync support

yacovm:
This commit adds sync support to the view:

- Whenever we get a message from the future from the leader of a view,
  we complain and then call Sync() from the view.
- The Sync triggers the controller to stop the current view,
  and start a new one with the view info returned from Sync().
- The Complain() now has an additional parameter which mentions
  whether the view changer is expected to close the view itself or not.
  This is to prevent the new view started after Sync() to be
  terminated right after it started by the view changer.

### DIFF
--- a/test/basic_test.go
+++ b/test/basic_test.go
@@ -6,6 +6,8 @@
 package test
 
 import (
+	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


test/test_app.go
```diff
@@ -117,13 +127,61 @@ func (a *App) AssembleProposal(metadata []byte, requests [][]byte) (nextProp typ
...
+	latestMD smartbftprotos.ViewMetadata
+	records  []*AppRecord
+}
+
+func (cb *committedBatches) add(record *AppRecord) {
+	cb.lock.Lock()
+	defer cb.lock.Unlock()
+
+	md := &smartbftprotos.ViewMetadata{}
+	proto.Unmarshal(record.Metadata, md)
```
golangcibot: Error return value of `proto.Unmarshal` is not checked (from `errcheck`)

yacovm: ugh, come on.. give me a break golangcibot :/

test/test_app.go
```diff
@@ -117,13 +127,63 @@ func (a *App) AssembleProposal(metadata []byte, requests [][]byte) (nextProp typ
...
-func (a *App) Deliver(proposal types.Proposal, signature []types.Signature) {
-	a.Delivered <- &AppRecord{
+func (a *App) Deliver(proposal types.Proposal, _ []types.Signature) {
+	record := &AppRecord{
 		Metadata: proposal.Metadata,
 		Batch:    BatchFromBytes(proposal.Payload),
 	}
+	a.Node.cb.add(record)
 	a.latestMD = &smartbftprotos.ViewMetadata{}
 	proto.Unmarshal(proposal.Metadata, a.latestMD)
```
HagarMeir, line 137: this also needs to be checked

yacovm: Sure, I'll address in a later PR since this isn't part of the change set.

test/basic_test.go
```diff
@@ -292,6 +294,60 @@ func TestLeaderForwarding(t *testing.T) {
...
+
+	for i := 1; i <= 10; i++ {
+		n0.Submit(Request{ID: fmt.Sprintf("%d", i), ClientID: "alice"})
+		<-n0.Delivered // Wait for leader to commit
+		<-n1.Delivered // Wait for follower to commit
+		<-n2.Delivered // Wait for follower to commit
+	}
+
+	n3.Connect()
+	// Submit a request to the follower which will trigger a timeout and then
```
HagarMeir, line 322: don't understand this comment

yacovm: this is a mistake, it should be sent to n1 or n2 or something, but i think i will just remove the comment itself in a subsequent PR.

HagarMeir (approved):

PR #290, opened 2019-08-25
Connect wal to test

tock-ibm:


test/test_app.go
```diff
@@ -212,3 +217,45 @@ func newNode(id uint64, network Network, testName string) *App {
 	app.Node = network[id]
 	return app
 }
+
+func initWAL(logger api.Logger, walDir string) (consensusWAL *wal.WriteAheadLogFile, walInitState [][]byte) {
```
yacovm, line 280: Is this logic equivalent to what is in the fabric repository? Can't we move this to a separate utility function in the WAL package?

test/basic_test.go
```diff
@@ -47,10 +53,14 @@ func TestRestartFollowers(t *testing.T) {
 	network := make(Network)
 	defer network.Shutdown()
 
-	n1 := newNode(1, network, t.Name())
-	n2 := newNode(2, network, t.Name())
-	n3 := newNode(3, network, t.Name())
-	n4 := newNode(4, network, t.Name())
+	testDir, err := ioutil.TempDir("", "test-restart-followers")
```
yacovm: you can use t.Name() in all of the tests instead explicitly specifying the test name, no?

test/test_app.go
tock-ibm: Yes it is, my plan is to do it a separate PR in which I also change the Fabric repo.

test/basic_test.go
tock-ibm: good idea!

yacovm (approved):

PR #292, opened 2019-08-25
Sync improvements

yacovm:


internal/bft/view.go
```diff
@@ -544,6 +541,69 @@ func (v *View) handlePrevSeqMessage(msgProposalSeq, sender uint64, m *protos.Mes
...
+		return
+	}
+
+	// To commit a block we need 2f + 1 votes.
+	// at least f+1 of them are honest and will broadcast
+	// their commits to votes to everyone including us.
+	// In each such a threshold of f+1 votes there is at least
+	// a single honest node that prepared for a proposal
+	// which we apparently missed.
+	_, f := computeQuorum(v.N)
```
HagarMeir, line 557: maybe read F at the start of the view

internal/bft/view.go
```diff
@@ -544,6 +541,69 @@ func (v *View) handlePrevSeqMessage(msgProposalSeq, sender uint64, m *protos.Mes
...
+		info := proposalInfo{
+			digest: vote.Digest,
+			view:   vote.View,
+			seq:    vote.Seq,
+		}
+		countsByVotes[info]++
+	}
+
+	// Check if there is a <digest, view, seq> that collected a threshold of votes,
+	// and that sequence is lower than out current sequence, or our view is different.
```
HagarMeir: with a sequence higher than our current sequence

internal/bft/view.go
```diff
@@ -544,6 +541,69 @@ func (v *View) handlePrevSeqMessage(msgProposalSeq, sender uint64, m *protos.Mes
...
+
+		// Disregard votes for past sequences for this view.
+		if vote.seq <= v.ProposalSequence && vote.view == v.Number {
+			continue
+		}
+
+		v.Logger.Warnf("Seen %d votes for digest %s in view %d, sequence %d but I am in view %d and seq %d",
+			count, vote.digest, vote.view, vote.seq, v.Number, v.ProposalSequence)
+		v.stop()
+		v.FailureDetector.Complain(false)
```
HagarMeir: why complain?

internal/bft/view_test.go
```diff
@@ -972,6 +972,132 @@ func TestViewPersisted(t *testing.T) {
 	}
 }
 
+func TestDiscoverDeliberateCensorship(t *testing.T) {
+	// Scenario: We test a use case in which the leader
+	// sends proposals to everyone but a single follower in order
+	// to censor it from the network.
+	// The node will eventually figure this out, and:
+	// 1) Try to lobby for a view change
```
HagarMeir: don't try to lobby for a view change if it doesn't complain

internal/bft/view_test.go
```diff
@@ -972,6 +972,132 @@ func TestViewPersisted(t *testing.T) {
...
+					syncCalled.Done()
+				})
+			}
+
+			view.Start()
+			defer view.Abort()
+
+			view.HandleMessage(2, makeCommit(testCase.firstMsgView, testCase.firstMsgSeq, "foo"))
+			view.HandleMessage(3, makeCommit(testCase.secondMsgView, testCase.secondMsgSeq, "foo"))
+			// Send some un-related messages, to ensure we processed the messages before this one.
```
HagarMeir, line 1082: why do you need this? didn't understand the comment

internal/bft/view_test.go
```diff
@@ -972,6 +972,132 @@ func TestViewPersisted(t *testing.T) {
...
+			for i := 0; i < 10*int(view.N)+2; i++ {
+				view.HandleMessage(2, prepare)
+			}
+
+			syncCalled.Wait()
+			if testCase.shouldSync {
```
HagarMeir: remove if we don't complain

HagarMeir (changes requested):

internal/bft/view.go
yacovm: This is a tradeoff betwen adding another field which would only be used for this function and computation time.

I think computation time here is negligible since it's basically integer arithmetic, and since it's only used here in this function it makes little sense to make a field for it.

internal/bft/view.go
yacovm: done

internal/bft/view.go
yacovm: done

internal/bft/view_test.go
yacovm: done

internal/bft/view_test.go
yacovm: done

internal/bft/view_test.go
yacovm: done

internal/bft/view.go
```diff
@@ -579,7 +579,7 @@ func (v *View) discoverIfSyncNeeded(sender uint64, m *protos.Message) {
 	}
 
 	// Check if there is a <digest, view, seq> that collected a threshold of votes,
-	// and that sequence is lower than out current sequence, or our view is different.
+	// and that sequence is higher than out current sequence, or our view is different.
```
HagarMeir: our (not out) current sequence

HagarMeir (approved):

PR #304, opened 2019-08-28
in flight proposal logic

HagarMeir:

### DIFF
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -6,6 +6,7 @@
 package bft
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/bft/viewchanger.go
```diff
@@ -499,6 +506,67 @@ func (v *ViewChanger) processViewDataMsg() {
 	}
 }
 
+func (v *ViewChanger) getViewDataMessages() []*protos.ViewData {
```
golangcibot, line 508: U1000: func `(*ViewChanger).getViewDataMessages` is unused (from `unused`)

HagarMeir: ok , will use this func in later commits

PR #305, opened 2019-08-28
Ignore heartbeats from leaders far ahead

yacovm:
This change makes the heartbeat monitor disregard heartbeats
from leaders that are far away as long as its view is functioning.

This is in order to eventually trigger a sync of a node that
is too far behind but no new proposals arrive and thus it lacks
the information to figure out it is far behind.

As part of the change set, the leader now also sends its proposal sequence
in the heart-beat.


internal/bft/controller.go
```diff
@@ -108,6 +109,8 @@ type Controller struct {
 	verificationSequence uint64
 
 	controllerDone sync.WaitGroup
+
+	ViewSequences *atomic.Value
```
HagarMeir, line 113: why is this in the controller?

yacovm: I need to keep the same reference somewhere for the: `(c *Consensus) proposalMaker() *algorithm.ProposalMaker` method to always construct a view with the same reference. 

I can put it either in the consensus or the controller, but in controller tests i need it anyway because they often have a view. 

So I decided to put it in the controller and just have the method above reference it.

test/basic_test.go
```diff
@@ -401,6 +405,85 @@ func TestCatchingUpWithSync(t *testing.T) {
...
+		if strings.Contains(entry.Message, "Leader's sequence is 10 and ours is 1") {
+			atomic.StoreUint32(&detectedSequenceGap, 1)
+		}
+		return nil
+	})).Sugar()
+
+	start := time.Now()
+	n0.heartbeatTime = make(chan time.Time, 1)
+	n0.heartbeatTime <- start
+	n3.heartbeatTime = make(chan time.Time, 1)
```
HagarMeir: duplicate?

pkg/consensus/consensus.go
```diff
@@ -112,6 +113,7 @@ func (c *Consensus) Start() {
 		Signer:           c.Signer,
 		RequestInspector: c.RequestInspector,
 		ViewChanger:      c.viewChanger,
+		ViewSequences:    &atomic.Value{},
```
HagarMeir, line 116: I think this should be in the consensus and not in the controller

internal/bft/heartbeatmonitor.go
```diff
@@ -199,11 +230,19 @@ func (hm *HeartbeatMonitor) leaderTick(now time.Time) {
 		return
 	}
 
-	hm.logger.Debugf("Sending heartbeat with view %d", hm.view)
+	var sequence uint64
+	vs := hm.viewSequences.Load()
+	if vs != nil && vs.(ViewSequence).ViewActive {
+		sequence = vs.(ViewSequence).ProposalSeq
+	} else {
+		hm.logger.Debugf("ViewSequence uninitialized")
```
HagarMeir: if the view is inactive this doesn't mean it is uninitialized

yacovm: Changed to: https://github.com/SmartBFT-Go/consensus/pull/305/commits/8287406ff2f3390b471abea448a13f306db6c9f2#diff-d964fc2921ff05a4258db85f3f3452aeR238

HagarMeir: if the view is inactive then we send heartbeat with sequence 0, maybe we should send the latests sequence anyway?

HagarMeir (approved):

PR #312, opened 2019-09-01
restart requests after view change completed

HagarMeir:


HagarMeir:
The TestMultiLeadersPartition is flaky

Issue #298, opened 2019-08-27
Create top level configuration object


yacovm:
Right now we have timeouts that are hardcoded or derived as a function of other timeouts.

We should have all configurable properties be configure-able via a top level configuration object.

C0rWin:
I guess that one should be in progress or we can consider it as done?

tock-ibm:
PR #314

tock-ibm:
Done

PR #314, opened 2019-09-01
Config object integration

tock-ibm:


pkg/consensus/consensus.go
```diff
@@ -59,16 +51,12 @@ func (c *Consensus) Deliver(proposal types.Proposal, signatures []types.Signatur
...
 
 func (c *Consensus) Start() {
-	// requestTimeout := 2 * c.BatchTimeout // Request timeout should be at least as batch timeout
-	opts := algorithm.PoolOptions{
-		QueueSize:         DefaultRequestPoolSize,
-		RequestTimeout:    algorithm.DefaultRequestTimeout / 100,
-		LeaderFwdTimeout:  algorithm.DefaultRequestTimeout / 5,
-		AutoRemoveTimeout: algorithm.DefaultRequestTimeout,
+	if c.Config.NumberOfNodes != len(c.Nodes()) {
+		// TODO return error
```
yacovm: Let's please remove the NumberOfNodes from the config. 
This is a plurality of configuration. 

Since we need to get the IDs of the nodes somehow, and currently we get them via the **Nodes()** API call - adding an additional configuration field that returns the number of nodes doesn't give any additional benefit and only makes the code more complex.

Given that we're going to need the **Nodes()** API call for reconfiguration,  it's probably here to stay - something I cannot say about the NumberOfNodes which is probably going to be removed once dynamic reconfiguration will be introduced. 

So let's please remove this field and stick with what we have now.

test/test_app.go
```diff
@@ -21,6 +21,20 @@ import (
...
+	RequestBatchMaxSize:       10,
+	RequestBatchMaxInterval:   time.Millisecond,
+	IncomingMessageBufferSize: 200,
+	RequestPoolSize:           40,
+	RequestTimeout:            500 * time.Millisecond,
+	RequestLeaderFwdTimeout:   2 * time.Second,
+	RequestAutoRemoveTimeout:  3 * time.Minute,
+	ViewChangeResendInterval:  time.Second,
+	ViewChangeTimeout:         1 * time.Minute,
+	LeaderHeartbeatTimeout:    4 * time.Second,
```
yacovm: This can be risky for tests in CI... we use logical clocks anyway for timeouts when we want it, so i suggest we go to a minute like before.

test/test_app.go
```diff
@@ -21,6 +21,20 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
+var FastConfig = consensus.Configuration{
```
yacovm: any reason this is exported?

yacovm (changes requested):
Please see comments, especially regarding the number of nodes.

pkg/consensus/consensus.go
yacovm: An alternative way is to also include the nodes in the configuration, and have the configuration be a function instead of a struct instance

tock-ibm: I favor the second alternative, have the library be explicitly configured with the identity of nodes.

test/test_app.go
tock-ibm: not really

pkg/consensus/consensus.go
yacovm: Then please add a function that returns a configuration object which contains the list of nodes (ids) and the size can be inferred from that list.

The function is what will be injected to the Consensus instead of the config object itself.

tock-ibm: that is not what I meant... I think since we still have no design for reconfiguration the best course of action is to remove NumberOfNodes and address reconfig when we have an issue for that.

test/test_app.go
tock-ibm: Done

test/test_app.go
tock-ibm: Done

yacovm (dismissed):

yacovm (approved):

tock-ibm:
#298

PR #315, opened 2019-09-01
add view change tests

HagarMeir:

### DIFF
--- a/test/basic_test.go
+++ b/test/basic_test.go
@@ -211,6 +211,23 @@ func TestMultiLeadersPartition(t *testing.T) {
 	n5 := newNode(5, network, t.Name(), testDir)
 	n6 := newNode(6, network, t.Name(), testDir)
 
+	start := time.Now()
+	n2.viewChangeTime = make(chan time.Time, 1)
+	n3.viewChangeTime = make(chan time.Time, 1)
+	n4.viewChangeTime = make(chan time.Time, 1)
+	n5.viewChangeTime = make(chan time.Time, 1)
+	n6.viewChangeTime = make(chan time.Time, 1)
+	n2.viewChangeTime <- start
+	n3.viewChangeTime <- start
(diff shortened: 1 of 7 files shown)


test/basic_test.go
```diff
@@ -320,6 +320,127 @@ func TestHeartbeatTimeoutCausesViewChange(t *testing.T) {
...
+		return nil
+	})).Sugar()
+
+	n0.Consensus.Start()
+	n1.Consensus.Start()
+
+	n0.Disconnect() // leader in partition
+	n1.Disconnect() // next leader in partition
+
+	viewChangeWG.Add(1)
```
yacovm, line 403: general rule of thumb - try to Add to the waitgroup right after its definition.
Who knows if you get a premature firing in a test and then you get a negative WG counter.

test/basic_test.go
```diff
@@ -320,6 +320,127 @@ func TestHeartbeatTimeoutCausesViewChange(t *testing.T) {
...
+	n4 := newNode(4, network, t.Name(), testDir)
+	n5 := newNode(5, network, t.Name(), testDir)
+	n6 := newNode(6, network, t.Name(), testDir)
+
+	start := time.Now()
+	n1.heartbeatTime = make(chan time.Time, 1)
+	n2.heartbeatTime = make(chan time.Time, 1)
+	n3.heartbeatTime = make(chan time.Time, 1)
+	n4.heartbeatTime = make(chan time.Time, 1)
+	n5.heartbeatTime = make(chan time.Time, 1)
```
yacovm: i believe you can fold this long initialization into something like:
for each on 6 nodes with n, and:
n.heartBeatTime = make(...)
n.viewChangeTime = make(...)
 add the start to both channels
setup n

yacovm (commented):
comments inline

yacovm (approved):

PR #326, opened 2019-09-05
blocking sync in view changer

HagarMeir:

### DIFF
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -629,6 +629,7 @@ func (v *ViewChanger) sync(targetSeq uint64) {
 		v.Synchronizer.Sync()
 		select { // wait for sync to return with expected info
 		case info := <-v.informChan:
+			// TODO what if view is bigger than expected?
 			if info.seq >= targetSeq {
 				return
 			}
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -636,6 +636,8 @@ func (v *ViewChanger) sync(targetSeq uint64) {
 		case now := <-v.Ticker:
 			v.lastTick = now
 			v.checkIfTimeout(now)
+		case change := <-v.startChangeChan:
+			v.startViewChange(change)
 		}
 	}
 }
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -34,6 +34,11 @@ type change struct {
 	stopView bool
 }
 
+type viewAndSeq struct {
+	view uint64
+	seq  uint64
+}
+
 type ViewChanger struct {
 	// Configuration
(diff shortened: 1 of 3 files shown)


internal/bft/viewchanger.go
```diff
@@ -607,14 +616,30 @@ func (v *ViewChanger) commitLastDecision(lastDecisionSequence uint64, lastDecisi
...
 	}
 	if md.LatestSequence > lastDecisionSequence+1 {
 		v.Logger.Panicf("Node %d has a checkpoint for sequence %d which is much greater than the last decision sequence %d", v.SelfID, md.LatestSequence, lastDecisionSequence)
 	}
 }
 
+func (v *ViewChanger) sync(targetSeq uint64) {
+	for {
+		v.Synchronizer.Sync()
+		select { // wait for sync to return with expected info
```
yacovm, line 630: I think you should also wait for startChangeChan for completeness, just to un-block anyone that is invoking StartViewChange.

yacovm (changes requested):
Please see comment in the code + maybe we should add a test in the **view_changer_test.go** for this,

HagarMeir:
I can add unit tests on Sunday

internal/bft/viewchanger_test.go
```diff
@@ -727,6 +727,121 @@ func TestCommitLastDecision(t *testing.T) {
...
+		Content: &protos.Message_NewView{
+			NewView: &protos.NewView{
+				SignedViewData: signed,
+			},
+		},
+	}
+
+	vc.Start(1)
+
+	synchronizerWG.Add(2)
```
yacovm, line 831: why is it safe to add the counter when the view changer operates in a different goroutine?

internal/bft/viewchanger_test.go
```diff
@@ -727,6 +727,121 @@ func TestCommitLastDecision(t *testing.T) {
...
+
+	vc.HandleMessage(1, msg)
```
yacovm, line 835: can you perhaps add some comments to give background what is the scenario, what is the expected behavior, etc?

internal/bft/viewchanger.go
```diff
@@ -607,14 +616,33 @@ func (v *ViewChanger) commitLastDecision(lastDecisionSequence uint64, lastDecisi
...
+		case info := <-v.informChan:
```
yacovm, line 631: if you read from **informChan** don't you also need to call **v.informNewView(info)** ?

HagarMeir: this is why I added a TODO later
`// TODO what if view is bigger than expected?`
it won't be simply calling the `informNewView`
and I will deal with this in a later PR

yacovm (approved):

PR #335, opened 2019-09-09
In flight

HagarMeir:


internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+			continue
+		}
+
+		if vd.InFlightProposal.Metadata == nil { // should have been validated earlier
+			panic(fmt.Sprintf("Node has a view data message where the in flight proposal metadata is nil"))
+		}
+
+		inFlightMetadata := &protos.ViewMetadata{}
+		if err := proto.Unmarshal(vd.InFlightProposal.Metadata, inFlightMetadata); err != nil { // should have been validated earlier
+			panic(fmt.Sprintf("Node was unable to unmarshal the in flight proposal metadata, error: %v", err))
```
yacovm: this is a message a remote node sends you, no? So why panic? Just skip it.

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+	for _, vd := range messages {
+		for _, possible := range possibleProposals {
+
+			if vd.InFlightProposal == nil {
+				possible.noArgument++
+				continue
+			}
+
+			// get the sequence
+			if vd.InFlightProposal.Metadata == nil {
```
yacovm: isn't this a message that a remote node sends you? Not sure you should panic here and in line 631

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+type possibleProposal struct {
+	proposal    *protos.Proposal
+	preprepared int
+	noArgument  int
+}
+
+func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, verifier api.Verifier) (ok, noInFlight bool, inFlightProposal *protos.Proposal) {
+	expectedSequence := MaxLastDecisionSequence(messages, quorum, N, verifier) + 1
+	possibleProposals := make([]*possibleProposal, 0)
+	noInFlightConut := 0
```
yacovm: noInFlightConut --> noInFlightCount or noInFlightCoconut

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+	for _, vd := range messages {
+
+		if vd.InFlightProposal == nil { // there is no in flight proposal here
+			noInFlightConut++
+			continue
+		}
+
+		if vd.InFlightProposal.Metadata == nil { // should have been validated earlier
```
yacovm, line 590: It's true, but this is an exported function, so either un-export it, or don't panic.

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
 	}
 }
 
+func (v *ViewChanger) getViewDataMessages() []*protos.ViewData {
+	num := len(v.viewDataMsgs.votes)
+	messages := make([]*protos.ViewData, 0)
+	for i := 0; i < num; i++ {
+		vote := <-v.viewDataMsgs.votes
+		vd := &protos.ViewData{}
+		if err := proto.Unmarshal(vote.GetViewData().RawViewData, vd); err != nil {
```
yacovm, line 557: general rule of thumb: never panic upon input from an external source.

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+			panic(fmt.Sprintf("Node was unable to unmarshal the in flight proposal metadata, error: %v", err))
+		}
+
+		if inFlightMetadata.LatestSequence != expectedSequence { // the in flight proposal sequence is not as expected
+			noInFlightConut++
+			continue
+		}
+
+		// find possible proposals
+		if inFlightMetadata.LatestSequence == expectedSequence { // this is the expected in flight proposal sequence
```
yacovm: once we passed: **if inFlightMetadata.LatestSequence != expectedSequence** above, then it also means if **inFlightMetadata.LatestSequence == expectedSequence**  so no reason to check once more.

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+			if vd.InFlightPrepared { // this proposal is prepared and so it is possible
```
yacovm: easier to check the inverse and then continue instead of having another indentation.

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+		}
+	}
+
+	// condition B holds
+	if noInFlightConut >= quorum { // there is a quorum of messages that support that there is no in flight proposal
+		return true, true, nil
+	}
+
+	// fill out info on all possible proposals
+	for _, vd := range messages {
```
yacovm: as a general rule of thumb - you just iterated over messages before, and did all kinds of validations.

Why not create a custom data structure (map, slice) that holds the distilled types you validated previously, so you don't need to validate again?

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+			noInFlightConut++
+			continue
+		}
+
+		// find possible proposals
+		if inFlightMetadata.LatestSequence == expectedSequence { // this is the expected in flight proposal sequence
+			if vd.InFlightPrepared { // this proposal is prepared and so it is possible
+				alreadyExists := false
+				for _, p := range possibleProposals {
+					if isSameProposals(p.proposal, vd.InFlightProposal) {
```
yacovm: how about creating a histogram (map) from a digest (string) to a list of proposals? 
Then you can easily pick the proposal that is pre-prepared, no?

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+
+	// condition A holds
+	if agreed != -1 {
+		return true, false, possibleProposals[agreed].proposal
+	}
+
+	return false, false, nil
+}
+
+func isSameProposals(p1, p2 *protos.Proposal) bool {
```
yacovm: this can be replaced by **proto.Equal()**

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+	if !isEqualBytes(p1.Metadata, p2.Metadata) {
+		return false
+	}
+	if !isEqualBytes(p1.Payload, p2.Payload) {
+		return false
+	}
+	return true
+}
+
+func isEqualBytes(a, b []byte) bool {
```
yacovm: can be replaced by **bytes.Equal**

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+	}
+	return true
+}
+
+func MaxLastDecisionSequence(messages []*protos.ViewData, quorum int, N uint64, verifier api.Verifier) uint64 {
+	max := uint64(0)
+	for _, vd := range messages {
+		err, seq := ValidateLastDecision(vd, quorum, N, verifier)
+		if err != nil {
+			panic(fmt.Sprintf("Node was unable to validate last decision in viewData message, error: %v", err))
```
yacovm, line 680: I don't think we should panic in exported functions such as these. We can return an error instead. 

In addition, this thing isn't fatal. It just means a specific VD is invalid, right? What if the node that sent that VD is byzantine and it just tries to make your crash?

internal/bft/viewchanger.go
```diff
@@ -512,20 +535,190 @@ func (v *ViewChanger) processViewDataMsg() {
...
+			return false
+		}
+	}
+	return true
+}
+
+func MaxLastDecisionSequence(messages []*protos.ViewData, quorum int, N uint64, verifier api.Verifier) uint64 {
+	max := uint64(0)
+	for _, vd := range messages {
+		err, seq := ValidateLastDecision(vd, quorum, N, verifier)
```
yacovm: Now that I look at it... we should really switch the order of the return arguments here. should have caught it in CR :/

internal/bft/viewchanger.go
```diff
@@ -657,4 +868,125 @@ func (v *ViewChanger) deliverDecision(proposal types.Proposal, signatures []type
...
+	// TODO maybePruneRevokedRequests
+	v.inFlightDecideChan <- struct{}{}
+}
+
+func (v *ViewChanger) Complain(viewNum uint64, stopView bool) {
+	v.Logger.Panicf("Node %d has complained while in the view for the in flight proposal", v.SelfID)
+}
+
+func (v *ViewChanger) Sync() {
+	// the in flight proposal view asked to sync
```
yacovm, line 958: can we sync in the view but not complain? only in **discoverIfSyncNeeded** but it shouldn't be able to get there at all if we abort the view inside **Decide**.

internal/bft/viewchanger.go
```diff
@@ -657,4 +868,125 @@ func (v *ViewChanger) deliverDecision(proposal types.Proposal, signatures []type
...
+		v.lastTick = now
+		v.checkIfTimeout(now)
+	}
+
+	v.inFlightView.Abort()
+}
+
+func (v *ViewChanger) Decide(proposal types.Proposal, signatures []types.Signature, requests []types.RequestInfo) {
+	v.Logger.Debugf("Delivering to app the last decision proposal %v", proposal)
+	v.Application.Deliver(proposal, signatures)
```
yacovm, line 942: Let's first abort the view, since it did its job, no?

yacovm (changes requested):
Good work overall, comments inside.

internal/bft/viewchanger.go
HagarMeir: Well the thing is that this message was validated earlier and if we got here we know it should be ok. If not then something is really wrong. 
There is also a comment: `// should have been validated earlier`

internal/bft/viewchanger.go
HagarMeir: see my reply above

internal/bft/viewchanger.go
HagarMeir: I exported it so I could test it. So I guess I'll return an error and deal with it later in the code.

HagarMeir: Or I could un-export it and check it in a separate test file that is a part of the same package?

internal/bft/viewchanger.go
HagarMeir: again was validate before reaching here so shouldn't happen

internal/bft/viewchanger.go
HagarMeir: same as before...

internal/bft/viewchanger.go
HagarMeir: why?

internal/bft/viewchanger.go
HagarMeir: Yes we sync here and not complain...
If this view complains then it will cause a panic...

```
func (v *ViewChanger) Complain(viewNum uint64, stopView bool) {
	v.Logger.Panicf("Node %d has complained while in the view for the in flight proposal", v.SelfID)
}
```

internal/bft/viewchanger.go
HagarMeir: In theory you are right
But the view is now busy with deciding so if I try to abort here the abort waits for the view to finish and we get a deadlock

internal/bft/viewchanger.go
yacovm: returning an error is best i think

internal/bft/viewchanger.go
yacovm: can you point me where?

internal/bft/viewchanger.go
yacovm: because in golang you return the error last, not first

internal/bft/viewchanger.go
yacovm: you can use **stop()**, no?

internal/bft/viewchanger.go
HagarMeir: ok will do

internal/bft/viewchanger.go
HagarMeir:
```
//viewData message
	if vd := m.GetViewData(); vd != nil {
		v.Logger.Debugf("Node %d is processing a view data message %s from %d", v.SelfID, MsgToString(m), sender)
		if !v.validateViewDataMsg(vd, sender) {
			return
		}
		v.viewDataMsgs.registerVote(sender, m)
		v.processViewDataMsg()
		return
	}
```

internal/bft/viewchanger.go
HagarMeir: alright

internal/bft/viewchanger.go
HagarMeir: to get a digest I need to convert to types.proposal
is it worth it?

PR #344, opened 2019-09-12
New view restore

HagarMeir:


internal/bft/state.go
```diff
@@ -69,11 +69,30 @@ func (ps *PersistedState) storeProposal(proposed *smartbftprotos.ProposedRecord)
 	ps.InFlightProposal.StoreProposal(proposalToStore)
 }
 
-func (ps *PersistedState) storePrepared(commitMsg *smartbftprotos.Message) {
+func (ps *PersistedState) storePrepared(commitMsg *protos.Message) {
 	cmt := commitMsg.GetCommit()
 	ps.InFlightProposal.StorePrepares(cmt.View, cmt.Seq)
 }
 
+func (ps *PersistedState) RestoreNewViewIfApplicable() (*types.ViewAndSeq, error) {
```
yacovm, line 77: The function name doesn't match its implementation... 

How about "LoadNewViewIfApplicable" that returns just the NewView protobuf message ?

pkg/consensus/consensus.go
```diff
@@ -128,6 +129,21 @@ func (c *Consensus) Start() {
 	c.viewChanger.Controller = c.controller
 	c.viewChanger.RequestsTimer = pool
 
+	viewSeq, err := c.state.RestoreNewViewIfApplicable()
+	if err != nil {
+		c.Logger.Warnf("Failed restoring new view, error: %v", err)
```
yacovm, line 134: Should be an Errorf I think. I'm also tempted to propose that we panic here, since we can only get an error if the WAL entry is corrupted, no?

pkg/consensus/consensus.go
```diff
@@ -128,6 +129,21 @@ func (c *Consensus) Start() {
...
+	}
+	if viewSeq == nil {
+		c.Logger.Debugf("No new view to restore")
+	} else {
+		// Check if metadata should be taken from the restored new view or from the application
+		if viewSeq.Seq >= c.Metadata.LatestSequence {
+			c.Logger.Debugf("Restoring from new view with view %d and seq %d, while application has view %d and seq %d", viewSeq.View, viewSeq.Seq, c.Metadata.ViewId, c.Metadata.LatestSequence)
+			c.viewChanger.Start(viewSeq.View)
```
yacovm, line 142: perhaps we can just have a local variable for the view number, and just override it in this *if* and preserve the place we start the view changer and controller below?

test/basic_test.go
```diff
@@ -554,6 +554,80 @@ func TestLeaderCatchingUpAfterViewChange(t *testing.T) {
...
+					n.app.heartbeatTime <- time.Now().Add(time.Second * time.Duration(2*i))
+				}
+			}
+		}
+	}()
+
+	viewChangeWG.Wait()
+	close(done)
+
+	n3.Restart()
```
yacovm, line 616: this is a follower centric test. should we perhaps make a similar one for a leader?

yacovm (approved):
Looks good, left some comments. 

Please address later on.

Issue #341, opened 2019-09-11
Batcher may pack duplicate requests in a batch


tock-ibm:
Say you have a remainder, when you call NextBatch(),
and then you call buildBatch() and ask for more items from the pool, and append them.
The prefix of the slice you get from the pool contains the same requests you already have in the remainder, i.e. in the currentBatch.

The only time a Batcher starts with remainder, is when BatchRemainder is called on it, and this happens in two places:
1. after pruning:
(c *Controller) maybePruneRevokedRequests()

2. if the Assembler returns a remainder:
AssembleProposal(metadata []byte, requests [][]byte) (nextProp bft.Proposal, remainder [][]byte)

tock-ibm:
I think we can get rid of the remainder concept: every time the batcher needs requests it will ask the pool for an array of requests that meets certain criteria (count, size).

yacovm:
What if we just make *NextRequests* in the request pool, remove requests and not just return them?

yacovm:
and, what you're basically saying is that since we anyway prune requests from request pool upon commit then we are going to prune anyway whatever we proposed, so what is left in the req. pool is effectively the reminder?

Issue #327, opened 2019-09-05
Make batcher honor total batch size in terms of megabytes as well as transaction count. 


yacovm:

tock-ibm:
Fabric has a configuration parameter:

message BatchSize {
    // Simply specified as number of messages for now, in the future
    // we may want to allow this to be specified by size in bytes
    uint32 max_message_count = 1;
    // The byte count of the serialized messages in a batch cannot
    // exceed this value.
    uint32 absolute_max_bytes = 2;
    // The byte count of the serialized messages in a batch should not
    // exceed this value.
    uint32 preferred_max_bytes = 3;
}

 we need to be compatible with that.

Issue #242, opened 2019-08-06
Make request pool use a scheduler


tock-ibm:
have a ticker emit time into a channel, and another thread read the time and scan events for deadline.

Possible advantage - better control of timing in tests, we can speed-up or slow down time as we wish. 

On the other hand this is more complicated, and has no clear benefit in production.

This can be applied to both RequestPool and Heartbeat monitor.

tock-ibm:
See #109 and discussion on #240

yacovm:
Update: Now the heartbeat monitor uses a time channel, and all we are left is to make the request pool use a scheduler.

PR #345, opened 2019-09-15
Batcher respect message count & size bytes

tock-ibm:
addresses both #341 and #327

### DIFF
--- a/internal/bft/batcher.go
+++ b/internal/bft/batcher.go
@@ -60,11 +60,6 @@ func (b *BatchBuilder) NextBatch() [][]byte {
 	}
 }
 
-// BatchRemainder sets the remainder of requests to be included in the next batch
-func (b *BatchBuilder) BatchRemainder(remainder [][]byte) {
-	// TODO remove method
-}
-
 // Close closes the close channel to stop NextBatch
 func (b *BatchBuilder) Close() {
 	b.closeLock.Lock()
@@ -92,10 +87,4 @@ func (b *BatchBuilder) Reset() {
 	b.closeLock.Lock()
 	defer b.closeLock.Unlock()
 	b.closeChan = make(chan struct{})
-}
-
-// PopRemainder returns the remainder and resets it
-func (b *BatchBuilder) PopRemainder() [][]byte {
-	// TODO remove method
-	return nil
 }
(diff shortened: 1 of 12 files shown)


internal/bft/sched_test.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 func TestSchedule(t *testing.T) {
+	t.Skip("flaky")
```
yacovm, line 18: I've never seen this fail in CI before...

tock-ibm: There is always a first time, sorry

tock-ibm: This object is not in use, anyway

tock-ibm: The flaky test will be fixed in #242 
It hangs in unit tests in the IDE

HagarMeir: it doesn't for me...weird

tock-ibm: When it is a timing issue, it can happen very rarely

internal/bft/requestpool.go
```diff
@@ -148,20 +150,29 @@ func (rp *Pool) Size() int {
...
+	count := minInt(rp.fifo.Len(), maxCount)
+	var totalSize uint64
+	batch = make([][]byte, 0, count)
 	var element = rp.fifo.Front()
-	for i := 0; i < m; i++ {
-		buff[i] = append(make([]byte, 0), element.Value.(*requestItem).request...)
+	for i := 0; i < count; i++ {
+		req := element.Value.(*requestItem).request
+		reqLen := uint64(len(req))
+		rp.logger.Debugf("i=%d, len(r)=%d, total=%d", i, len(req), totalSize)
```
yacovm, line 166: I'm not sure this is the ideal place for this logging. I think ideally we will have a logging just before each place we return from the function (2 places total)

internal/bft/batcher.go
```diff
@@ -12,64 +12,54 @@ import (
...
 		case <-b.closeChan:
 			return nil
 		case <-timeout:
-			return b.buildBatch(remainderOccupied, currBatch)
+			currBatch, _ = b.pool.NextRequests(b.maxMsgCount, b.maxSizeBytes)
+			return currBatch
 		default:
-			if b.pool.Size() >= b.batchSize-remainderOccupied {
-				return b.buildBatch(remainderOccupied, currBatch)
+			time.Sleep(10 * time.Millisecond)
```
yacovm, line 52: i think long term we should restructure this whole thing so we won't have to sleep at all, but have some kind of synchronization that eliminates sleeping.

yacovm (approved):

internal/bft/requestpool.go
tock-ibm: Correct. I added this during debug and forgot to remove. I'll push an update.

internal/bft/batcher.go
tock-ibm: I agree.

PR #349, opened 2019-09-16
view change message in WAL

HagarMeir:


test/basic_test.go
```diff
@@ -637,6 +637,126 @@ func TestRestartAfterViewChangeAndRestoreNewView(t *testing.T) {
...
+			case <-time.After(time.Millisecond * 10):
+				for _, n := range network {
+					n.app.heartbeatTime <- time.Now().Add(time.Second * time.Duration(2*i))
+					n.app.viewChangeTime <- time.Now().Add(time.Second * time.Duration(2*i))
+				}
+			}
+		}
+	}()
+
+	viewChangeWG.Wait()
```
yacovm, line 734: Why not disconnect the node in the place you call viewChangeWG.Done ?

test/basic_test.go
```diff
@@ -637,6 +637,126 @@ func TestRestartAfterViewChangeAndRestoreNewView(t *testing.T) {
...
+		if strings.Contains(entry.Message, "Node 3 got a view change timeout, the current view is 1") {
+			viewChangeTimeoutWG.Done()
+		}
+		if strings.Contains(entry.Message, "ViewChanged, the new view is 2") {
+			viewChangeFinishWG.Done()
+		}
+		return nil
+	})).Sugar()
+
+	start := time.Now()
```
yacovm, line 701: the section below will ideally be right after network creation

internal/bft/viewchanger.go
```diff
@@ -316,18 +319,29 @@ func (v *ViewChanger) startViewChange(change *change) {
...
-	if len(v.viewChangeMsgs.voted) >= v.quorum-1 && v.nextView > v.currView { // send view data
+	if (len(v.viewChangeMsgs.voted) >= v.quorum-1 && v.nextView > v.currView) || restore { // send view data
+		msgToSave := &protos.SavedMessage{
+			Content: &protos.SavedMessage_ViewChange{
+				ViewChange: &protos.ViewChange{
+					NextView: v.currView,
+				},
+			},
+		}
+		if err := v.State.Save(msgToSave); err != nil {
```
yacovm, line 341: do we always want to save it again even if it is the last message? Shouldn't we skip saving it if the restore flag is on?

yacovm (approved):

yacovm:
Looks generally good, a couple of comments to think about.

test/basic_test.go
HagarMeir: because that place is called again
that is why I even add to this wait group later
`viewChangeWG.Add(2) // to avoid negative WaitGroup counter, n2 will send the view data again after restoring the view change`

test/basic_test.go
HagarMeir: don't see why that matters but ok

Issue #350, opened 2019-09-18
Improve sync to bring new-view


tock-ibm:
If a leader gets partitioned and the partition heals, but after the view change there were no transactions, the current Synchronizer.Sync() will not inform the node on the new view.

The core of the problem is because our state-transfer is partial - compared to pbft and bft-smart - we do not bring the most recent part of the "log" (in the general sense), which in our case is the WAL, in the sync. We only bring what was decided and logged in the app's log (i.e. get the blocks form Fabric).

Therefore, one way to solve problem, and possibly other problems that we did not tackle yet, is to improve the sync process in general. We leave the Syc() implemented in Fabric as is, to do catch up on decided blocks.

On top of that, we add a "state-transfer" protocol in the library, that would bring to the requesting node the part of the state in peer nodes that is not provided in Synchronizer.Sync() - that is, information about view changes. 

In particular, the "ViewChange" & "NewView" messages written to the WAL are of interest to this protocol.

An outline of the protocol is:
- The controller sync() would call the Fabric (dependency Synchronizer) Sync() that catches up decided blocks and brings the last decision. 
- Then (or in parallel), it could call a "State-transfer-request" (for lack of a better term). Peers respond to that by sending in a "State-transfer-response" message, that includes some description of the latest messages written to the WAL (from memory, actually). 
- Nodes don't have to send the propose+prepare or commit in full, as that is delivered by Sync(); they have to send only an indication of the View+Sequence an the phase they are in; e.g.: <'NewView', v, s> or <'Prepare', v, s> or <'Commit', v, s>, etc.
- The requesting node can then wait for f+1 identical responses and make progress from there.

yacovm:
So, all we need is to get the NewView message. 
I suggest that the "state transfer request" will simply contain a view sequence number, and the response will contain a **NewView** message if the view number is greater than the number of the request, or simply an acknowledgement that the view number is the same.  

We also don't need to wait for f+1 identical responses when we get a response containing a **NewView** message because it's signed.

HagarMeir:
Right now the **NewView** message is not actually written to the WAL, just the view and sequence agreed on, but it is easy to add it if we want to check signatures.

And remember that even if one node sent a **NewView** message it doesn't mean that it is the latest view. You need at least `f+1` to know what is the latest you are expecting.

yacovm:
> Right now the NewView message is not actually written to the WAL, just the view and sequence agreed on, but it is easy to add it if we want to check signatures.

[...]

PR #358, opened 2019-09-22
state transfer on sync

HagarMeir:


internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+			s.responses.registerVote(msg.sender, msg.Message)
+			if viewAndSeq := s.collectLogic(); viewAndSeq != nil {
+				s.Logger.Infof("Node %d collected a valid state: view - %d and seq - %d", s.SelfID, viewAndSeq.View, viewAndSeq.Seq)
+				timer.Stop()
+				return viewAndSeq
+			}
+		}
+	}
+
+	return nil
```
golangcibot: unreachable: unreachable code (from `govet`)

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (s *StateCollector) collectLogic() *types.ViewAndSeq {
+	if len(s.responses.voted) > s.f {
+		votesMap := make(map[types.ViewAndSeq]uint64, 0)
```
golangcibot: S1019: should use make(map[types.ViewAndSeq]uint64) instead (from `gosimple`)

HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/350

internal/bft/controller.go
```diff
@@ -425,10 +444,27 @@ func (c *Controller) sync() {
...
+		Content: &protos.Message_StateTransferRequest{
+			StateTransferRequest: &protos.StateTransferRequest{},
+		},
+	}
+	c.Comm.BroadcastConsensus(msg)
+	response := c.Collector.CollectStateResponses()
+	if response != nil {
+		if response.View > md.ViewId && response.Seq == md.LatestSequence {
+			c.Logger.Infof("The collected state is with view %d and sequence %d", response.View, response.Seq)
+			view = response.View
```
yacovm, line 462: in case of a view change, the nodes that changed the view - write the view change metadata to WAL. 
This node, however - doesn't. 
Shouldn't we write it to WAL as well as soon as we successfully detect the view has increased without the seqNum being increased, to mimic what the rest of the cluster does?

internal/bft/controller_test.go
```diff
@@ -488,6 +488,14 @@ func TestSyncInform(t *testing.T) {
 	controllerMock := &mocks.ViewController{}
 	controllerMock.On("AbortView", mock.Anything)
 
+	collector := bft.StateCollector{
+		SelfID:         0,
+		N:              4,
+		Logger:         log,
+		CollectTimeout: 10 * time.Millisecond,
```
yacovm: I'd increase this to a few seconds to avoid flakes

internal/bft/controller.go
```diff
@@ -425,10 +444,27 @@ func (c *Controller) sync() {
...
+
+	view := md.ViewId
+
+	msg := &protos.Message{
+		Content: &protos.Message_StateTransferRequest{
+			StateTransferRequest: &protos.StateTransferRequest{},
+		},
+	}
+	c.Comm.BroadcastConsensus(msg)
+	response := c.Collector.CollectStateResponses()
```
yacovm, line 458: the CollectStateResponses drains the message channel before receiving, isn't that a problem? 
You might get a response as you drain the channel in tests.
I think that you should actually drain the channel as you aggregate the responses, at the end of CollectStateResponses and not at its beginning.

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+	case s.incMsgs <- msg:
+	default: // if incMsgs is full do nothing
+		s.Logger.Debugf("Node %d reached default in handling state response: %v", s.SelfID, msg)
+	}
+}
+
+// CollectStateResponses return a valid response or nil if reached timeout
+func (s *StateCollector) CollectStateResponses() *types.ViewAndSeq {
+	// drain message channel
+	for len(s.incMsgs) > 0 {
```
yacovm, line 70: this block can moved to the end of the method, since you are anyway protected by the votes and no one can vote more than once.

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+		<-s.incMsgs
+	}
+
+	s.responses.clear(s.N)
+
+	timer := time.NewTimer(s.CollectTimeout)
```
yacovm, line 79: you can just defer timer.Stop() here

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+		case <-s.stopChan:
+			timer.Stop()
+			return nil
+		case <-timer.C:
+			s.Logger.Infof("Node %d reached the state collector timeout", s.SelfID)
+			return nil
+		case msg := <-s.incMsgs:
+			s.Logger.Debugf("Node %d collected a response: %v", s.SelfID, msg)
+			s.responses.registerVote(msg.sender, msg.Message)
+			if viewAndSeq := s.collectLogic(); viewAndSeq != nil {
```
yacovm: "collectLogic" can perhaps be renamed to something like: "collectedQuorumOfX" when X is either votes, or view attestations, etc.

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+				return viewAndSeq
+			}
+		}
+	}
+
+	return nil
+}
+
+func (s *StateCollector) collectLogic() *types.ViewAndSeq {
+	if len(s.responses.voted) > s.f {
```
yacovm: you can check the NOT of this and just early return nil, and then reduce indentation for all the function.

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+		votesMap := make(map[types.ViewAndSeq]uint64, 0)
+		num := len(s.responses.votes)
+		for i := 0; i < num; i++ {
+			vote := <-s.responses.votes
+			viewAndSeq := types.ViewAndSeq{
+				View: vote.GetStateTransferResponse().ViewNum,
```
yacovm: nil check missing ?

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+		return message.GetStateTransferResponse() != nil
+	}
+	s.responses = &voteSet{
+		validVote: acceptResponse,
+	}
+	s.responses.clear(s.N)
+}
+
+// HandleMessage handle messages addressed to the state collector
+func (s *StateCollector) HandleMessage(sender uint64, m *protos.Message) {
```
yacovm, line 52: let's check the message type here, and that the inner type isn't nil - just in case.

internal/bft/statecollector.go
```diff
@@ -0,0 +1,143 @@
...
+			vote := <-s.responses.votes
+			viewAndSeq := types.ViewAndSeq{
+				View: vote.GetStateTransferResponse().ViewNum,
+				Seq:  vote.GetStateTransferResponse().Sequence,
+			}
+			s.Logger.Debugf("Node %d collected a responses with view - %d and seq - %d", s.SelfID, viewAndSeq.View, viewAndSeq.Seq)
+			s.responses.votes <- vote
+			if count, exist := votesMap[viewAndSeq]; exist {
+				count++
+				votesMap[viewAndSeq] = count
```
yacovm: I think that votesMap[viewAndSeq]++ should work... no?

pkg/consensus/config.go
```diff
@@ -53,6 +53,9 @@ type Configuration struct {
 	// LeaderHeartbeatCount is the number of heartbeats per LeaderHeartbeatTimeout that the leader should emit.
 	// The heartbeat-interval is equal to: LeaderHeartbeatTimeout/LeaderHeartbeatCount.
 	LeaderHeartbeatCount int
+
+	// CollectTimeout is the interval after which the node stops listening to StateTransferResponse messages.
```
yacovm: Perhaps rephrase to: "The timeout that it takes to collect information about view metadata from remote nodes." ?

internal/bft/statecollector_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/SmartBFT-Go/consensus/internal/bft"
+	"github.com/SmartBFT-Go/consensus/pkg/types"
+	protos "github.com/SmartBFT-Go/consensus/smartbftprotos"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func TestBasicTimeout(t *testing.T) {
```
yacovm, line 22: put parallel ?

internal/bft/statecollector_test.go
```diff
@@ -0,0 +1,165 @@
...
+	collector := &bft.StateCollector{
+		SelfID:         0,
+		N:              4,
+		Logger:         log,
+		CollectTimeout: 10 * time.Millisecond,
+	}
+
+	collector.Start()
+
+	collector.CollectStateResponses()
```
yacovm: check the return value is nil?

yacovm (changes requested):
Did a quick pass, looks good ! 

Some comments below.

internal/bft/statecollector_test.go
```diff
@@ -0,0 +1,165 @@
...
+func TestBasicTimeout(t *testing.T) {
+	basicLog, err := zap.NewDevelopment()
+	assert.NoError(t, err)
+	log := basicLog.Sugar()
+
+	collector := &bft.StateCollector{
+		SelfID:         0,
+		N:              4,
+		Logger:         log,
+		CollectTimeout: 10 * time.Millisecond,
```
yacovm: if we put t.Parallel() in the test, you can increase the timeout to like 1 second, or a few hundred millis, to avoid flakes.

internal/bft/controller.go
HagarMeir: good point

internal/bft/controller_test.go
HagarMeir: Why? I want there to be a timeout...the faster the better

internal/bft/controller.go
HagarMeir: I don't want to handle old responses
in uint tests I wait to see that collection has started by using the log

pkg/consensus/config.go
HagarMeir: added it

yacovm (changes requested):
one last issue :)

yacovm (approved):

yacovm (approved):

PR #366, opened 2019-09-24
Leader heartbeat response

tock-ibm:


internal/bft/controller.go
```diff
@@ -230,6 +230,13 @@ func (c *Controller) OnHeartbeatTimeout(view uint64, leaderID uint64) {
 	c.FailureDetector.Complain(c.getCurrentViewNumber(), true)
 }
 
+// OnHeartbeatResponseSync is called when a leader receives enough heartbeat responses that indicate its view is
+// outdated. Called by the HeartbeatMonitor goroutine.
+func (c *Controller) OnHeartbeatResponseSync(view uint64) {
```
yacovm, line 235: Cant we please not add yet another method to the controller which only calls a function that is already exported to other components? Why not just log the below message inside the HBM and just call the Sync instead?

internal/bft/heartbeatmonitor.go
```diff
@@ -34,47 +42,52 @@ type roleChange struct {
 	follower Role
 }
 
+// senderSet is a set of node IDs
+type senderSet map[uint64]bool
+
+// heartbeatResponseCollector is a map from view ID to a set of senders.
+type heartbeatResponseCollector map[uint64]senderSet
```
yacovm, line 49: Here is an alternative idea that I think is superior because it limits the amount of messages you save to **N**.

Instead of saving a **set** per **view number**, you save **view numbers** per **sender**.
Whenever a message is received, you just overwrite the value (view number) in the key (sender ID), and then check the length of the map to see if it is larger than **f+1**. 

Since you anyway ignore messages that are in your own view or in a past view - you end up only storing votes for higher views. After collecting f+1 votes you know that there is at least 1 honest node that is in a higher view than you. 

Of course, if we want to be really strict, we can just iterate in some order over the mapping and create an in-memory histogram and count f+1 identical votes upon every reception of a message. 
This is slightly more expensive in computation but this is really negligible.

internal/bft/heartbeatmonitor.go
```diff
@@ -34,47 +42,52 @@ type roleChange struct {
...
-func NewHeartbeatMonitor(
-	scheduler <-chan time.Time,
-	logger api.Logger,
-	heartbeatTimeout time.Duration,
-	heartbeatCount int,
-	comm Comm,
-	handler HeartbeatTimeoutHandler,
-	viewSequences *atomic.Value,
-) *HeartbeatMonitor {
+func NewHeartbeatMonitor(scheduler <-chan time.Time, logger api.Logger, heartbeatTimeout time.Duration, heartbeatCount int, comm Comm, numberOfNodes uint64, handler HeartbeatTimeoutHandler, syncHandler HeartbeatResponseSyncHandler, viewSequences *atomic.Value) *HeartbeatMonitor {
```
yacovm, line 76: can we please collapse it back to have 1 parameter in each line?

test/test_app.go
```diff
@@ -24,7 +24,7 @@ import (
 var fastConfig = consensus.Configuration{
 	RequestBatchMaxCount:      10,
 	RequestBatchMaxBytes:      10 * 1024 * 1024,
-	RequestBatchMaxInterval:   time.Millisecond,
+	RequestBatchMaxInterval:   10 * time.Millisecond,
```
yacovm, line 27: How is this related to the change?

test/basic_test.go
```diff
@@ -194,6 +194,84 @@ func TestAfterDecisionLeaderInPartition(t *testing.T) {
...
+	defer network.Shutdown()
+
+	testDir, err := ioutil.TempDir("", t.Name())
+	assert.NoErrorf(t, err, "generate temporary test dir")
+	defer os.RemoveAll(testDir)
+
+	logConfig := zap.NewDevelopmentConfig()
+	logger, _ := logConfig.Build()
+	logger = logger.With(zap.String("t", t.Name())).With(zap.String("id", "TEST"))
+	sugaredLogger := logger.Sugar()
```
yacovm, line 210: why do you need this logger? if you want to log test messages you can just use t.Log(), but i don't think we should use this. Please remove.

test/basic_test.go
```diff
@@ -194,6 +194,84 @@ func TestAfterDecisionLeaderInPartition(t *testing.T) {
...
+	assert.Equal(t, data1Tx4, data2)
+	assert.Equal(t, data2, data3)
+
+	n0.Connect() // partition heals, leader should eventually sync and deliver
+	sugaredLogger.Infof("Connected n0")
+
+	data0 = <-n0.Delivered
+	assert.Equal(t, data1Tx3, data0)
+
+	data0 = <-n0.Delivered
```
yacovm, line 271: can't we simplify the test? For instance, we only need 1 extra block, not 2.

test/test_app.go
```diff
@@ -53,6 +53,10 @@ type App struct {
 	logger         *zap.SugaredLogger
 }
 
+func (a *App) SetLogLevel(level zapcore.Level) {
```
yacovm, line 56: I don't think this is used.... if not, please remove

internal/bft/heartbeatmonitor_test.go
```diff
@@ -284,27 +275,171 @@ func TestHeartbeatMonitorLeaderAndFollower(t *testing.T) {
...
 	handler1.AssertCalled(t, "OnHeartbeatTimeout", uint64(12), uint64(2))
 	handler1.AssertNumberOfCalls(t, "OnHeartbeatTimeout", 1)
 }
 
+func TestHeartbeatResponse(t *testing.T) {
+	basicLog, err := zap.NewDevelopment()
+	assert.NoError(t, err)
+	log := basicLog.Sugar()
+
+	t.Run("leader", func(t *testing.T) {
```
yacovm, line 296: these 2 tests don't share anything, so I think it's best to extract them into 2 different functions entirely and no make sub-tests.

yacovm (changes requested):
Did a quick pass

internal/bft/heartbeatmonitor.go
```diff
@@ -145,33 +158,84 @@ func (hm *HeartbeatMonitor) ChangeRole(follower Role, view uint64, leaderID uint
...
-	if hbMsg.View != hm.view {
-		hm.logger.Infof("Heartbeat view is different than monitor view, ignoring; view: %d, sender: %d, msg: %v", hm.leaderID, sender, msg)
+	if hm.view >= hbr.View {
+		hm.logger.Debugf("Monitor view: %d >= HeartBeatResponse, ignoring; sender: %d, msg: %v", hm.view, sender, hbr)
 		return
 	}
 
-	if hm.viewActiveButBehindLeader(hbMsg) {
+	hm.logger.Debugf("Received HeartBeatResponse, msg: %v; from %d", hbr, sender)
+	senders, exists := hm.hbRespCollector[hbr.View]
```
yacovm, line 207: What if a malicious node sends you infinite responses? You will end up saving them all, no?

internal/bft/controller.go
tock-ibm: Good idea. 
I need an interface for this so I suggest to unify both calls to the controller in one interface:

```
type HeartbeatEventHandler interface {
   OnHeartbeatTimeout(view uint64, leaderID uint64)
   Sync()
}
```

This is implemented by the controller, and also simplifies the construction in consensus.

>

internal/bft/heartbeatmonitor.go
tock-ibm: We need to detect f+1 **identical** votes, i.e. on the same view, that is for sure.
The problem with the current implementation is that indeed a rogue node can force the old leader to store a lot of entries in the map, e.g. it just sends responses with ever increasing view numbers. This must be prevented, I agree.

internal/bft/heartbeatmonitor.go
tock-ibm: sure

test/test_app.go
tock-ibm: not really related, but makes sense nevertheless

internal/bft/heartbeatmonitor.go
tock-ibm: I agree, responded to that above as well.
I will change that.

internal/bft/heartbeatmonitor.go
yacovm: Why do we need to detect identical votes? All we need to know is that there is an honest node that its view surpasses you, no?

internal/bft/heartbeatmonitor_test.go
tock-ibm: done

internal/bft/heartbeatmonitor.go
yacovm: @HagarMeir  ?

HagarMeir: I don't see why they need to be identical and not simply higher.
And to avoid the problem Yoav described we should save a view per sender and not a set per view.

yacovm: Let's think of it this way - if the leader is sending its view number, in theory - the follower can respond with a bit saying - "my view is higher than yours". 
In such a case, we would need to wait for **f+1** identical bits, right?
So here we are just representing this in a different way (sending the view number itself) but the safety holds in an equivalent way.

test/basic_test.go
tock-ibm: done

test/basic_test.go
tock-ibm: done

test/test_app.go
tock-ibm: done

tock-ibm:
pushed a new one

PR #367, opened 2019-09-25
Leader nacks after review

tock-ibm:
Addressed code reviews from #366 and rebased on latest master


internal/bft/heartbeatmonitor.go
```diff
@@ -198,30 +201,44 @@ func (hm *HeartbeatMonitor) handleHeartBeatResponse(sender uint64, hbr *smartbft
...
 	}
 
-	_, f := computeQuorum(hm.numberOfNodes)
-	if len(senders) >= f+1 {
-		hm.logger.Debugf("Calling HeartBeatResponse Sync, view: %d", hbr.View)
-		hm.syncHandler.OnHeartbeatResponseSync(hbr.View)
+	var doSync bool
+	histogram := make(heartbeatResponseHistogram)
+	for _, view := range hm.hbRespCollector {
+		numVotes := histogram[view] // if does not exist, returns 0
```
yacovm: histogram[view]++ works just fine, I think.

internal/bft/heartbeatmonitor.go
```diff
@@ -198,30 +201,44 @@ func (hm *HeartbeatMonitor) handleHeartBeatResponse(sender uint64, hbr *smartbft
...
+		numVotes++
+		histogram[view] = numVotes
+		if numVotes >= f+1 {
```
yacovm: I don't understand why we really need **f+1** identical view numbers ? 
Why can't we just observe **f+1** attestations that the view number is higher than ours? 
Clearly, if we have Byzantine nodes - they can just all send the same view number that an honest node sent us, so every attack that is possible when we count views higher than our view, is also possible when we count identical views, no?

internal/bft/heartbeatmonitor.go
tock-ibm: It does work, changed.

internal/bft/heartbeatmonitor.go
tock-ibm: This argument is convincing.

internal/bft/heartbeatmonitor.go
```diff
@@ -34,25 +38,34 @@ type roleChange struct {
 	follower Role
 }
 
+// heartbeatResponseCollector is a map from node ID to view number, and hold the last response from each node.
+type heartbeatResponseCollector map[uint64]uint64
+
+// heartbeatResponseHistogram stores how many responses exist per view.
+type heartbeatResponseHistogram map[uint64]int
```
golangcibot: `heartbeatResponseHistogram` is unused (from `deadcode`)

internal/bft/heartbeatmonitor.go
```diff
@@ -145,33 +158,82 @@ func (hm *HeartbeatMonitor) ChangeRole(follower Role, view uint64, leaderID uint
...
+	if hm.view >= hbr.View {
+		hm.logger.Debugf("Monitor view: %d >= HeartBeatResponse, ignoring; sender: %d, msg: %v", hm.view, sender, hbr)
 		return
 	}
 
-	hm.logger.Debugf("Received heartbeat from %d, last heartbeat was %v ago", sender, hm.lastTick.Sub(hm.lastHeartbeat))
-	hm.lastHeartbeat = hm.lastTick
+	hm.logger.Debugf("Received HeartBeatResponse, msg: %v; from %d", hbr, sender)
+	prevVote, exists := hm.hbRespCollector[sender]
+	if exists && prevVote == hbr.View {
```
yacovm: Why do we care if it exists? If we reached this point then we know that hbr.View > hm.View, so why not always re-write it?

test/basic_test.go
```diff
@@ -224,6 +224,67 @@ func TestAfterDecisionLeaderInPartition(t *testing.T) {
...
+	n3 := newNode(3, network, t.Name(), testDir)
+
+	n0.Consensus.Start()
+	n1.Consensus.Start()
+	n2.Consensus.Start()
+	n3.Consensus.Start()
+
+	n0.Submit(Request{ID: "1", ClientID: "alice"}) // submit to leader
+
+	data0 := <-n0.Delivered
```
yacovm, line 249: Is this a sanity check? Why not just disconnect the leader upfront without doing transactions prior to the disconnect?

internal/bft/heartbeatmonitor.go
```diff
@@ -145,33 +158,82 @@ func (hm *HeartbeatMonitor) ChangeRole(follower Role, view uint64, leaderID uint
...
+	case *smartbftprotos.Message_HeartBeatResponse:
+		hm.handleHeartBeatResponse(sender, msg.GetHeartBeatResponse())
+	default:
+		hm.logger.Warnf("Unexpected message type, ignoring")
+	}
+}
+
+func (hm *HeartbeatMonitor) handleHeartBeat(sender uint64, hb *smartbftprotos.HeartBeat) {
+	if sender != hm.leaderID || hb.View != hm.view {
+		hm.logger.Infof("Heartbeat is unexpected, ignoring; Monitor: leader=%d, view=%d; HB: sender: %d, view: %d, seq: %d",
```
yacovm: I see that you have combined the 2 cases and now it logs the same thing. The previous code distinguishes between the 2 cases and it logs messages that IMO are more user friendly.

yacovm (changes requested):
some small nits

internal/bft/heartbeatmonitor.go
tock-ibm: done

internal/bft/heartbeatmonitor.go
tock-ibm: The problem is that these two conditions may overlap. Before it didn't matter, because we were not sending Nacks, just returning early. Now, apart from logging issue, the code is more compact. I will change the log message though, so that it is clear which conditions are violated.

test/basic_test.go
tock-ibm: Yes, I want to do this after a couple of TXs are already past the first leader.

yacovm (approved):

PR #378, opened 2019-10-17
Pool size bytes

HagarMeir:


internal/bft/requestpool.go
```diff
@@ -243,10 +252,12 @@ func (rp *Pool) RemoveRequest(requestInfo types.RequestInfo) error {
 		return fmt.Errorf(errStr)
 	}
 
-	return rp.deleteRequest(element, requestInfo)
+	rp.deleteRequest(element, requestInfo)
+	rp.sizeBytes -= uint64(len(element.Value.(*requestItem).request))
```
yacovm, line 256: this should be inside deleteRequest in my opinion.

yacovm (approved):

Issue #309, opened 2019-08-29
don't stop accepting new requests when stopping timers


HagarMeir:
Today when the view changer calls StopTimers it not only stops the timers but also prevents new requests from getting in to the pool.
StopTimers should only stop the timers and keep the pool open for new requests.

yacovm:
...

PR #381, opened 2019-10-23
allow submit of new requests when stopped timers

HagarMeir:

### DIFF
--- a/internal/bft/requestpool.go
+++ b/internal/bft/requestpool.go
@@ -51,6 +51,7 @@ type Pool struct {
 	semaphore      *semaphore.Weighted
 	existMap       map[types.RequestInfo]*list.Element
 	timeoutHandler RequestTimeoutHandler
+	closed         bool
 	stopped        bool
 	submittedChan  chan struct{}
 	sizeBytes      uint64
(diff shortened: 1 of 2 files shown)


HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/309

internal/bft/requestpool.go
```diff
@@ -128,6 +129,10 @@ func (rp *Pool) Submit(request []byte) error {
 		rp.options.RequestTimeout,
 		func() { rp.onRequestTO(reqCopy, reqInfo) },
 	)
+	if rp.stopped {
+		rp.logger.Debugf("pool stopped, submitting with a stopped timer, request: %s", reqInfo)
+		to.Stop()
+	}
```
yacovm, line 135: this is executed under the lock, so just check if it's stopped right after acquiring the lock, and return early.

internal/bft/requestpool.go
```diff
@@ -51,6 +51,7 @@ type Pool struct {
 	semaphore      *semaphore.Weighted
 	existMap       map[types.RequestInfo]*list.Element
 	timeoutHandler RequestTimeoutHandler
+	closed         bool
 	stopped        bool
```
yacovm, line 55: Do we need the stopped now ?

yacovm (changes requested):
see comments

internal/bft/requestpool.go
HagarMeir: we don't return....we submit the request with a stopped timer...

internal/bft/requestpool.go
HagarMeir: yes...to know during the submit that the pool is stopped and to stop the new request timer

yacovm (approved):

Issue #241, opened 2019-08-06
Heartbeat monitor improvement: skip heartbeats when data is send by leader


tock-ibm:
On leader, When the leader sends a message by broadcast, skip pending HB.
On follower, when a message is received from the leader, extend HB timeout accordingly.

This saves some minute amount of bandwidth when blocks are decided on a regular basis.

tock-ibm:
See #109

PR #386, opened 2019-10-27
heartbeat on view messages

HagarMeir:

### DIFF
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -54,6 +54,7 @@ type RequestPool interface {
 type LeaderMonitor interface {
 	ChangeRole(role Role, view uint64, leaderID uint64)
 	ProcessMsg(sender uint64, msg *protos.Message)
+	HeartbeatWasSent()
 	Close()
 }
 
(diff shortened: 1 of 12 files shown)


HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/241

pkg/consensus/consensus.go
```diff
@@ -209,7 +208,7 @@ func (c *Consensus) BroadcastConsensus(m *protos.Message) {
 func (c *Consensus) proposalMaker() *algorithm.ProposalMaker {
 	return &algorithm.ProposalMaker{
 		State:           c.state,
-		Comm:            c,
+		Comm:            c.controller,
```
yacovm, line 205: The BroadcastConsensus method can now be removed.

internal/bft/controller.go
```diff
@@ -280,6 +284,29 @@ func (c *Controller) respondToStateTransferRequest(sender uint64) {
 	c.Comm.SendConsensus(sender, msg)
 }
 
+func (c *Controller) convertViewMessageToHeartbeat(m *protos.Message) *protos.Message {
+	var view, seq uint64
+	switch m.GetContent().(type) {
+	case *protos.Message_PrePrepare:
```
yacovm: we have methods that extract the seq and view number in util.go, you can use them.

yacovm (changes requested):
a few nits + that thing i sent via slack

Issue #380, opened 2019-10-23
Remainder is not being used


HagarMeir:
Consider changing the API and removing the remainder from the assembler

PR #394, opened 2019-11-03
document dependencies

HagarMeir:

### DIFF
--- a/pkg/api/dependencies.go
+++ b/pkg/api/dependencies.go
@@ -10,45 +10,73 @@ import (
 	protos "github.com/SmartBFT-Go/consensus/smartbftprotos"
 )
 
+// Application delivers the committed proposal and corresponding signatures.
 type Application interface {
+	// Deliver delivers the given proposal and signatures.
+	// After the call returns we assume that this proposal is stored in persistent memory.
 	Deliver(proposal bft.Proposal, signature []bft.Signature)
 }
 
+// Comm enables the communications between the nodes.
 type Comm interface {
+	// SendConsensus sends the consensus protocol related message m to the node with id targetID.
 	SendConsensus(targetID uint64, m *protos.Message)
+	// SendTransaction sends the given client's request to the node with id targetID.
 	SendTransaction(targetID uint64, request []byte)
+	// Nodes returns a set of ids of participating nodes.
 	Nodes() []uint64
 }
 
+// Assembler creates proposals.
 type Assembler interface {
+	// AssembleProposal creates a proposal which includes
+	// the given requests (when permitting) and metadata.
 	AssembleProposal(metadata []byte, requests [][]byte) (nextProp bft.Proposal, remainder [][]byte)
 }
 
+// WriteAheadLog is a write ahead log.
 type WriteAheadLog interface {
+	// Append appends a data item to the end of the WAL
+	// and indicate whether this entry is a truncation point.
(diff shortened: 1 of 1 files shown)


HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/300

pkg/api/dependencies.go
```diff
@@ -10,45 +10,73 @@ import (
...
 	SendTransaction(targetID uint64, request []byte)
+	// Nodes returns a set of ids of participating nodes.
 	Nodes() []uint64
 }
 
+// Assembler creates proposals.
 type Assembler interface {
+	// AssembleProposal creates a proposal which includes
+	// the given requests (when permitting) and metadata.
 	AssembleProposal(metadata []byte, requests [][]byte) (nextProp bft.Proposal, remainder [][]byte)
```
C0rWin, line 34: IMO need also elaborate on reminder.

pkg/api/dependencies.go
```diff
@@ -10,45 +10,73 @@ import (
 	protos "github.com/SmartBFT-Go/consensus/smartbftprotos"
 )
 
+// Application delivers the committed proposal and corresponding signatures.
```
C0rWin: Is it indeed committed proposal? I was under impression this is a callback interface which triggers only after replicas have mutual consent.

pkg/api/dependencies.go
```diff
@@ -10,45 +10,73 @@ import (
...
 
+// Verifier validates data and verifies signatures.
 type Verifier interface {
+	// VerifyProposal verifies the given proposal and returns the included requests' info.
 	VerifyProposal(proposal bft.Proposal) ([]bft.RequestInfo, error)
+	// VerifyRequest verifies the given request and returns its info.
 	VerifyRequest(val []byte) (bft.RequestInfo, error)
+	// VerifyConsenterSig verifies the signature for the given proposal.
 	VerifyConsenterSig(signature bft.Signature, prop bft.Proposal) error
+	// 	VerifySignature verifies the signature.
```
C0rWin: A nit, but can you please fix indentation?

pkg/api/dependencies.go
HagarMeir: Actually the remainder is not used.
https://github.com/SmartBFT-Go/consensus/issues/380
Maybe we should remove it completely.

pkg/api/dependencies.go
HagarMeir: This is called after replicas as agreed...after all three phases of consensus. 
Is "committed" not clear enough?

C0rWin: Committed there?

pkg/api/dependencies.go
C0rWin: If not in use, we definitely need to get rid of it.

pkg/api/dependencies.go
HagarMeir: Committed = agreed on
How do you suggest I change it to make it more clear?

yacovm: Consented should be a better term

pkg/api/dependencies.go
yacovm: :+1:

HagarMeir: OK then that will be my next task

yacovm (approved):

Issue #370, opened 2019-09-26
Join view change after f+1 complained - Discussion


HagarMeir:
Currently we follow the protocol suggested by BFT SMaRt which states that a node will join the view change after receiving f+1 view change messages. 
This may cause more false alarms and more view changes, however this is supposed to help speed up the view change process.
We should think if we want instead to join the view change only after the node itself complains or it got a quorum of view change messages.

PR #399, opened 2019-11-04
join view change after quorum

HagarMeir:


HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/370

yacovm (changes requested):
I think be default it should be turned off in the tests

yacovm (approved):

Issue #403, opened 2019-11-13
Stream-line Configuration - validation, types


tock-ibm:
Validate() on Configuration
except:

```
nodes := c.Comm.Nodes()
    for _, val := range nodes {
        if val == 0 {
            return errors.Errorf("node id 0 is not permitted")
        }
    }
```
which will stay in `Consensus.Start()`

Check that self ID is within "nodes".

Swap the definition of Config.RequestTimeout vs. RequestLeaderFwdTimeout

RequestTimeout, intuitively, should be the 2nd TO, where as the short, first one, should be RequestLeaderFwdTimeout

All integer parameters that are >=0 by nature (e.g. byte size) move to uint64

tock-ibm:
Done

PR #404, opened 2019-11-17
Config streamline

tock-ibm:
See #403


pkg/consensus/consensus.go
```diff
@@ -228,65 +228,16 @@ func (c *Consensus) proposalMaker() *algorithm.ProposalMaker {
...
-	if c.Config.RequestTimeout > c.Config.RequestLeaderFwdTimeout {
-		return errors.Errorf("RequestTimeout is bigger than RequestLeaderFwdTimeout")
-	}
-	if c.Config.RequestLeaderFwdTimeout > c.Config.RequestAutoRemoveTimeout {
-		return errors.Errorf("RequestLeaderFwdTimeout is bigger than RequestAutoRemoveTimeout")
-	}
-	if c.Config.ViewChangeResendInterval > c.Config.ViewChangeTimeout {
-		return errors.Errorf("ViewChangeResendInterval is bigger than ViewChangeTimeout")
-	}
+
```
HagarMeir, line 254: you also said that we should check if SelfID is within Nodes

HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/403

pkg/consensus/consensus.go
tock-ibm: True, I'll add that to the issue so that I don't forget. It will be in a future commit - there are a couple more that I plan on - the types, that check, etc.

internal/bft/requestpool.go
```diff
@@ -66,18 +66,18 @@ type requestItem struct {
 // PoolOptions is the pool configuration
 type PoolOptions struct {
 	QueueSize         int64
-	RequestTimeout    time.Duration
-	LeaderFwdTimeout  time.Duration
+	ForwardTimeout    time.Duration
```
yacovm, line 69: the name of the timeout is supposed to be defined by the operation that is timed out, not by what happens afterwards.

pkg/consensus/config.go
```diff
@@ -88,3 +92,63 @@ var DefaultConfig = Configuration{
...
+	if !(c.RequestBatchMaxCount > 0) {
+		return errors.Errorf("RequestBatchMaxCount should be greater than zero")
+	}
+	if !(c.RequestBatchMaxBytes > 0) {
+		return errors.Errorf("RequestBatchMaxBytes should be greater than zero")
+	}
+	if !(c.RequestBatchMaxInterval > 0) {
+		return errors.Errorf("RequestBatchMaxInterval should be greater than zero")
+	}
+	if !(c.IncomingMessageBufferSize > 0) {
```
yacovm, line 110: is this an unsigned it? if so it's always > 0, so why not just check equality with 0 ? same for the other uints.

pkg/consensus/config.go
```diff
@@ -88,3 +92,63 @@ var DefaultConfig = Configuration{
...
+		return errors.Errorf("IncomingMessageBufferSize should be greater than zero")
+	}
+	if !(c.RequestPoolSize > 0) {
+		return errors.Errorf("RequestPoolSize should be greater than zero")
+	}
+	if !(c.RequestForwardTimeout > 0) {
+		return errors.Errorf("RequestForwardTimeout should be greater than zero")
+	}
+	if !(c.RequestComplainTimeout > 0) {
```
yacovm, line 119: why not just check if it's <= 0 ?

yacovm (changes requested):
comments inline

internal/bft/requestpool.go
HagarMeir: Maybe change it to:
RequestTimeout  and SecondTimeout
or something like this

pkg/consensus/config.go
HagarMeir: This is my original code...
Feel free to change it as you wish

internal/bft/requestpool.go
tock-ibm: I don't think there is a rule like that, as long as it is consistent and intuitive, which wasn't the case.

pkg/consensus/config.go
tock-ibm: I id not change the conditions, they are all the same as they were. This format is preferable because it is "sign proof", it will work even if someone decides to change the type back to int.

pkg/consensus/config.go
tock-ibm: same as above

internal/bft/requestpool.go
yacovm: This is a configuration, right? So the name of the field should denote what is the thing that times out, and not what happens as a result of a timeout. 
What happens as a result of a timeout, is a detail of the protocol and the user shouldn't be even aware of it.

yacovm (approved):

PR #408, opened 2019-12-25
call abort view only after controller starts

HagarMeir:

### DIFF
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -100,6 +100,8 @@ type ViewChanger struct {
 	stopOnce sync.Once
 	stopChan chan struct{}
 	vcDone   sync.WaitGroup
+
+	ControllerStartedWG *sync.WaitGroup
 }
 
 // Start the view changer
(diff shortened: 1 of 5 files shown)


internal/bft/viewchanger.go
```diff
@@ -100,6 +100,8 @@ type ViewChanger struct {
 	stopOnce sync.Once
 	stopChan chan struct{}
 	vcDone   sync.WaitGroup
+
+	ControllerStartedWG *sync.WaitGroup
```
yacovm: No need to make it a pointer. 
Here we can just make it a non pointer and then you don't need to initialize it in the tests, since it's initialized implicitly.

pkg/consensus/consensus.go
```diff
@@ -186,6 +187,11 @@ func (c *Consensus) Start() error {
 		}
 	}
 
+	controllerStartedWG := sync.WaitGroup{}
+	controllerStartedWG.Add(1)
+	c.controller.StartedWG = &controllerStartedWG
```
yacovm: let's just make the controllerStartedWG a non pointer as a field of the view changer, and then we can just put the pointer of the waitgroup in the controller field.

yacovm (changes requested):
Just a small nit

yacovm (approved):

Issue #409, opened 2020-01-07
Initiate sync if a follower is constantly a single sequence behind the leader in the monitor


yacovm:
If we're a follower that is only 1 proposal behind, we will never sync autonomously to avoid  false negatives that involve the leader being ahead of us and starting the next proposal, thus syncing needlessly.

However a side effect of this, is that a follower can remain stuck a single proposal behind the rest of the cluster for a very long time, until there are new requests.

To that end I think that the monitor should track the period of time that it is a single sequence behind the leader, and if it stays this way for above a certain threshold (say, 15 seconds) then it should initiate a sync similarly to how it acts when the leader reports of a higher sequence.

PR #410, opened 2020-01-13
sync when follower is behind

HagarMeir:

### DIFF
--- a/internal/bft/heartbeatmonitor.go
+++ b/internal/bft/heartbeatmonitor.go
@@ -68,10 +68,14 @@ type HeartbeatMonitor struct {
 	viewSequences       *atomic.Value
 	sentHeartbeat       chan struct{}
 	artificialHeartbeat chan incMsg
+	behindSeq           uint64
+	behindCounter       uint64
+	behindNumOfTicks    uint64
+	followerBehind      bool
 }
 
 // NewHeartbeatMonitor creates a new HeartbeatMonitor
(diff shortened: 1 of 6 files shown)


HagarMeir:
https://github.com/SmartBFT-Go/consensus/issues/409

internal/bft/heartbeatmonitor.go
```diff
@@ -358,6 +370,16 @@ func (hm *HeartbeatMonitor) followerTick(now time.Time) {
 	}
 
 	hm.logger.Debugf("Last heartbeat from %d was %v ago", hm.leaderID, delta)
+
+	if hm.followerBehind {
```
yacovm: can we check the complement condition and return from the function instead of creating an indented block?

internal/bft/heartbeatmonitor.go
```diff
@@ -68,10 +68,14 @@ type HeartbeatMonitor struct {
 	viewSequences       *atomic.Value
 	sentHeartbeat       chan struct{}
 	artificialHeartbeat chan incMsg
+	behindSeq           uint64
+	behindCounter       uint64
+	behindNumOfTicks    uint64
```
yacovm: `behindNumOfTicks` sounds a bit like `behindCounter`, can we simply call it `numOfTicksBehindBeforeSyncing` since we are anyway injecting it into the field?

yacovm (approved):

PR #417, opened 2020-01-28
reconfig on deliver

HagarMeir:


pkg/consensus/consensus.go
```diff
@@ -63,10 +67,15 @@ func (c *Consensus) Complain(viewNum uint64, stopView bool) {
 	c.viewChanger.StartViewChange(viewNum, stopView)
 }
 
-func (c *Consensus) Deliver(proposal types.Proposal, signatures []types.Signature) {
+func (c *Consensus) Deliver(proposal types.Proposal, signatures []types.Signature) bool {
```
yacovm: Can't we return the reconfig object here? The name of the method is `deliver` so it's not clear from its name, why it returns a boolean and what the boolean means.

test/reconfig_test.go
```diff
@@ -54,17 +54,132 @@ func TestBasicReconfig(t *testing.T) {
...
 	for i := 0; i < numberOfNodes; i++ {
 		d := <-nodes[i].Delivered
 		data = append(data, d)
 	}
 	for i := 0; i < numberOfNodes-1; i++ {
 		assert.Equal(t, data[i], data[i+1])
 	}
+}
+
+func TestBasicAddNode(t *testing.T) {
```
yacovm: looks good

test/test_app.go
```diff
@@ -73,14 +73,26 @@ func (a *App) Submit(req Request) {
 // Sync synchronizes and returns the latest decision
 func (a *App) Sync() types.SyncResponse {
 	records := a.Node.cb.readAll(*a.latestMD)
+	reconfigSync := types.ReconfigSync{InReplicatedDecisions: false}
```
yacovm, line 76: false is the default value, we can just initialize it with `var reconfigSync types.ReconfigSync`

pkg/consensus/consensus.go
```diff
@@ -166,8 +167,14 @@ func (c *Consensus) reconfig(reconfig types.Reconfig) {
 	c.collector.Stop()
 
 	c.Config = reconfig.CurrentConfig
-	if err := c.ValidateConfiguration(); err != nil {
-		c.Logger.Panicf("Configuration is invalid, error: %v", err)
+	if err := c.ValidateConfiguration(reconfig.CurrentNodes); err != nil {
+		if strings.Contains(err.Error(), "Nodes does not contain the SelfID") {
+			c.close()
+			c.Logger.Infof("Closing consensus")
```
yacovm: let's explain explicitly why we're closing.

pkg/consensus/consensus.go
```diff
@@ -256,26 +263,25 @@ func (c *Consensus) proposalMaker() *algorithm.ProposalMaker {
...
 	if err := c.Config.Validate(); err != nil {
 		return errors.Wrap(err, "bad configuration")
 	}
 
-	nodes := c.Comm.Nodes()
 	nodeSet := make(map[uint64]bool)
 	for _, val := range nodes {
 		if val == 0 {
-			return errors.Errorf("Comm.Nodes() contains node id 0 which is not permitted, nodes: %v", nodes)
+			return errors.Errorf("Nodes contains node id 0 which is not permitted, nodes: %v", nodes)
```
yacovm: errors start with lowercase

test/test_app.go
HagarMeir: It is more clear to me this way, if you don't mind

yacovm (approved):

PR #427, opened 2020-02-13
fixing view change

HagarMeir:

### DIFF
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -509,7 +509,7 @@ func (v *ViewChanger) checkLastDecision(svd *protos.SignedViewData, sender uint6
 		return false, 0
 	}
 
-	v.Logger.Debugf("Node %d got %s from %d, the last decision seq is %d and this node's current sequence %d", v.SelfID, signedViewDataToString(svd), sender, lastDecisionMD.LatestSequence, mySequence)
+	v.Logger.Debugf("Node %d got %s from %d, the last decision seq is %d and this node's current sequence is %d", v.SelfID, signedViewDataToString(svd), sender, lastDecisionMD.LatestSequence, mySequence)
 
 	if lastDecisionMD.LatestSequence > mySequence+1 { // this is a decision in the future, ignoring since the node might not have the right configuration to validate
 		v.Logger.Debugf("Node %d got %s from %d, but the last decision seq %d is greater than this node's current sequence %d", v.SelfID, signedViewDataToString(svd), sender, lastDecisionMD.LatestSequence, mySequence)
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -470,10 +470,14 @@ func (v *ViewChanger) validateViewDataMsg(svd *protos.SignedViewData, sender uin
 		return false
 	}
 
+	v.Logger.Debugf("Node %d got %s from %d, and it passed the last decision check", v.SelfID, signedViewDataToString(svd), sender)
+
 	if err := ValidateInFlight(vd.InFlightProposal, lastDecisionSequence); err != nil {
 		v.Logger.Warnf("Node %d got %v from %d, but the in flight proposal is invalid, reason: %v", v.SelfID, signedViewDataToString(svd), sender, err)
 		return false
 	}
+
+	v.Logger.Debugf("Node %d got %s from %d, and the in flight proposal is valid", v.SelfID, signedViewDataToString(svd), sender)
 
 	return true
 }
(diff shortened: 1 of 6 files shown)


internal/bft/viewchanger.go
```diff
@@ -448,42 +448,149 @@ func (v *ViewChanger) getInFlight(lastDecision *protos.Proposal) *protos.Proposa
...
 	}
-	v.Logger.Debugf("Node %d got %s from %d, the last decision with sequence %d is valid", v.SelfID, viewDataToString(vd), sender, lastSequence)
-	if err := ValidateInFlight(rvd.InFlightProposal, lastSequence); err != nil {
-		v.Logger.Warnf("Node %d got %v from %d, but the in flight proposal is invalid, reason: %v", v.SelfID, viewDataToString(vd), sender, err)
+
+	if err := ValidateInFlight(vd.InFlightProposal, lastDecisionSequence); err != nil {
+		v.Logger.Warnf("Node %d got %v from %d, but the in flight proposal is invalid, reason: %v", v.SelfID, signedViewDataToString(svd), sender, err)
 		return false
 	}
-	v.Logger.Debugf("Node %d got %s from %d, the in flight proposal is valid", v.SelfID, viewDataToString(vd), sender)
```
yacovm, line 483: This line doesn't exist in the new version, and is pretty valuable

internal/bft/viewchanger.go
```diff
@@ -448,42 +448,149 @@ func (v *ViewChanger) getInFlight(lastDecision *protos.Proposal) *protos.Proposa
...
+		// compare the last decision itself
+		if !proto.Equal(vd.LastDecision, myLastDecision) {
+			v.Logger.Warnf("Node %d got %s from %d, they are at the same sequence but the last decisions are not equal", v.SelfID, signedViewDataToString(svd), sender)
+			return false, 0
+		}
+
+		return true, lastDecisionMD.LatestSequence
+	}
+
+	if lastDecisionMD.LatestSequence == mySequence+1 {
```
yacovm: perhaps check the complement condition and fail fast (return false) instead of indenting this whole block

internal/bft/viewchanger.go
```diff
@@ -448,42 +448,149 @@ func (v *ViewChanger) getInFlight(lastDecision *protos.Proposal) *protos.Proposa
...
+			return false, 0
+		}
+
+		proposal := types.Proposal{
+			Header:               vd.LastDecision.Header,
+			Metadata:             vd.LastDecision.Metadata,
+			Payload:              vd.LastDecision.Payload,
+			VerificationSequence: int64(vd.LastDecision.VerificationSequence),
+		}
+		signatures := make([]types.Signature, 0)
```
yacovm: var signatures []types.Signature

internal/bft/viewchanger.go
```diff
@@ -552,37 +659,43 @@ func ValidateInFlight(inFlightProposal *protos.Proposal, lastSequence uint64) er
...
+	if err != nil {
+		v.Logger.Panicf("Node %d checked the in flight and it got an error: %v", v.SelfID, err)
+	}
+	if !ok {
+		v.Logger.Debugf("Node %d checked the in flight and it was invalid", v.SelfID)
+		return
+	}
+	v.Logger.Debugf("Node %d checked the in flight and it was valid", v.SelfID)
+	// create the new view message
+	signedMsgs := make([]*protos.SignedViewData, 0)
```
yacovm: var signedMsgs []*protos.SignedViewData

internal/bft/viewchanger.go
```diff
@@ -552,37 +659,43 @@ func ValidateInFlight(inFlightProposal *protos.Proposal, lastSequence uint64) er
...
+	myMsg := v.prepareViewDataMsg()                      // since it might have changed by now
+	signedMsgs = append(signedMsgs, myMsg.GetViewData()) // leader's message will always be the first
+	close(v.viewDataMsgs.votes)
+	for vote := range v.viewDataMsgs.votes {
```
yacovm, line 687: I think we should never have "processed" our own view data message in the first place, but let's take care of it in a subsequent PR

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
 func maxLastDecisionSequence(messages []*protos.ViewData, quorum int, N uint64, verifier api.Verifier) uint64 {
```
yacovm: these quorum, N, verifier are not used anymore. please remove them?

yacovm: these quorum, N, verifier are not used anymore. please remove them?

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
...
 	max := uint64(0)
 	for _, vd := range messages {
```
yacovm, line 837: I might be paranoid but I think we should add a parameter to this method that is the current sequence, and then just check if there is a sequence that is our seq + 1 and return it

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
...
+			v.Logger.Warnf("Node %d is processing newView message, but the last decision of %s is not set", v.SelfID, signedViewDataToString(svd))
+			return false, false, false
 		}
 
-		if err := ValidateInFlight(vd.InFlightProposal, lastSequence); err != nil {
-			v.Logger.Warnf("Node %d is processing newView message, but the in flight in %s is invalid, reason: %v", v.SelfID, viewDataToString(svd), err)
+		if vd.LastDecision.Metadata == nil { // this is a genesis proposal
+			if mySequence > 0 {
+				// can't validate the signature since I am ahead
+				if err := ValidateInFlight(vd.InFlightProposal, 0); err != nil {
```
yacovm, line 886: if we're ahead then why don't we also check that the content of the in flight proposal is equal to our latest decision?

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
...
+			v.Synchronizer.Sync() // TODO check if I manged to sync to latest decision, revalidate new view, and join the other nodes
+			return true, true, false
 		}
-		if !ok {
-			v.Logger.Debugf("The check of the in flight proposal by node %d did not pass", v.SelfID)
-			return
+
+		if lastDecisionMD.LatestSequence < mySequence { // this is a decision in the past
+			// can't validate the signature since I am ahead
+			if err := ValidateInFlight(vd.InFlightProposal, lastDecisionMD.LatestSequence); err != nil {
```
yacovm, line 922: why don't we also check the content of the in flight?

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
...
+			signatures := make([]types.Signature, 0)
+			for _, sig := range vd.LastDecisionSignatures {
+				signature := types.Signature{
+					ID:    sig.Signer,
+					Value: sig.Value,
+					Msg:   sig.Msg,
+				}
+				signatures = append(signatures, signature)
+			}
+			v.deliverDecision(proposal, signatures)
```
yacovm, line 974: Why do we blindly deliver this without calling **v.Verifier.VerifyProposal()** beforehand?

internal/bft/viewchanger.go
```diff
@@ -714,189 +827,231 @@ func CheckInFlight(messages []*protos.ViewData, f int, quorum int, N uint64, ver
...
-				return
-			}
-		case <-v.stopChan:
-		case now := <-v.Ticker:
-			v.lastTick = now
-			v.checkIfTimeout(now)
-		case change := <-v.startChangeChan:
-			v.startViewChange(change)
+	if !noInFlight {
+		if !v.commitInFlightProposal(inFlightProposal) {
```
yacovm: this can be collapsed into a single if condition

test/basic_test.go
```diff
@@ -1187,6 +1217,69 @@ func TestGradualStart(t *testing.T) {
 
 }
 
+func TestReconfigAndViewChange(t *testing.T) {
```
yacovm, line 1220: can you please add some comments to explain what is going on in this test?

internal/bft/viewchanger.go
HagarMeir: added a comment to clarify

internal/bft/viewchanger.go
HagarMeir: We check the in flight proposals later
We don't compare it now since the in flight proposal might be different then the last decision and the view data message is still considered valid

internal/bft/viewchanger.go
HagarMeir: same as the last comment

internal/bft/viewchanger.go
HagarMeir: we don't...we deliver only after calling `ValidateLastDecision`

yacovm (approved):

PR #453, opened 2021-01-27
get leader id

HagarMeir:

### DIFF
--- a/pkg/consensus/consensus.go
+++ b/pkg/consensus/consensus.go
@@ -61,7 +61,7 @@ type Consensus struct {
 
 	reconfigChan chan types.Reconfig
 
-	running bool
+	running uint64
 }
 
 func (c *Consensus) Complain(viewNum uint64, stopView bool) {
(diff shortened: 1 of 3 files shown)


pkg/consensus/consensus.go
```diff
@@ -90,6 +92,16 @@ func (c *Consensus) Sync() types.SyncResponse {
 	return syncResponse
 }
 
+// GetLeaderID returns the current leader ID or zero if Consensus is not running
+func (c *Consensus) GetLeaderID() uint64 {
+	c.consensusLock.RLock()
+	defer c.consensusLock.RUnlock()
+	if !c.running {
```
yacovm: Instead of locking, let's just use an atomic uint32 flag.

The `c.leaderID()` acquires a read lock and you have already acquired a read lock here.

This will make us enter a re-entrant lock deadlock by:
T1: RLock
T2: WLOCK - waits for acquiring lock
T1: Rlock - waits for all readers to finish

yacovm (approved):

PR #481, opened 2021-12-23
Delete in-flight when syncing past the in-flight sequence

yacovm:
In case a node syncs past its last recorded in-flight proposal, it needs to delete the in-flight because the rest of the nodes
have progressed further than this in-flight proposal.

Hence it will never need to initialize a view to consent on the in-flight proposal since during the sync it has already committed this proposal, and even if there are some nodes that still think this response is in-flight, the current node can make them commit this proposal as it has corresponding signatures from the decision of this proposal.

Signed-off-by: Yacov Manevich <yacovm@il.ibm.com>

PR #482, opened 2021-12-23
gave the node a chance to keep up with the others

pfi79:


internal/bft/view.go
```diff
@@ -202,6 +204,21 @@ func (v *View) processMsg(sender uint64, m *protos.Message) {
 	// This message is either for this proposal or the next one (we might be behind the rest)
 	if msgProposalSeq != v.ProposalSequence && msgProposalSeq != v.ProposalSequence+1 {
 		v.Logger.Warnf("%d got message from %d with sequence %d but our sequence is %d", v.SelfID, sender, msgProposalSeq, v.ProposalSequence)
+
+		if msgProposalSeq > v.ProposalSequence {
+			fv, ok := v.futuredMsg[msgProposalSeq]
```
yacovm: How do you keep this map from growing indefinitely? If you commit rate is considerably slower than the commit rate of the rest of the nodes, this map will grow and grow until you run out of memory.

pfi79: If two comits from the future come, it means we are very very late. And the chance didn't come true (discoverIfSyncNeeded).

internal/bft/view.go
```diff
@@ -851,6 +868,34 @@ func (v *View) startNextSeq() {
 	v.commits = v.nextCommits
 	tmpVotes.clear(v.N)
 	v.nextCommits = tmpVotes
+
+	if fv, ok := v.futuredMsg[v.ProposalSequence+1]; ok {
```
yacovm: I don't think this is the right place to put this. If you are saving messages that arrived too early then you should probably enqueue them in a channel and then dispatch them in the main loop: 

```
	for {
               // check if we have any messages that arrived late
              ....
              //
		select {
		case <-v.abortChan:
			return
		case msg := <-v.incMsgs:
			v.processMsg(msg.sender, msg.Message)
		default:
			v.doPhase()
		}
	}
```

but not here.

pfi79: @yacovm Throw them in v.incMsgs? 
Can I call v.HandleMessage(sender, m)?

pfi79: But there may be a problem when v.incMsgs is fully populated.

yacovm: I would do something like that:

```
diff --git a/internal/bft/view.go b/internal/bft/view.go
index d7fd4f1..1b04be4 100644
--- a/internal/bft/view.go
+++ b/internal/bft/view.go
@@ -72,6 +72,7 @@ type View struct {
        Phase              Phase
        InMsgQSize         int
        // Runtime
+       futureMsgs         map[uint64]*voteSet
        lastVotedProposalByID map[uint64]protos.Commit
        incMsgs               chan *incMsg
        myProposalSig         *types.Signature
@@ -104,6 +105,7 @@ type View struct {
 
 // Start starts the view
 func (v *View) Start() {
+       v.futureMsgs = make(map[uint64]*voteSet)
        v.stopOnce = sync.Once{}
        v.incMsgs = make(chan *incMsg, v.InMsgQSize)
        v.abortChan = make(chan struct{})
@@ -193,6 +195,12 @@ func (v *View) processMsg(sender uint64, m *protos.Message) {
                return
        }
 
+       if msgProposalSeq > v.ProposalSequence + 1 && len(v.futureMsgs) < 10 && msgProposalSeq < v.ProposalSequence + 10{
+               v.maybeInitializeFutureMsgVoteSetForSeq(msgProposalSeq)
+               v.futureMsgs[msgProposalSeq].registerVote(sender, m)
+               return
+       }
+
        if msgProposalSeq == v.ProposalSequence-1 && v.ProposalSequence > 0 {
                v.handlePrevSeqMessage(msgProposalSeq, sender, m)
                return
@@ -238,6 +246,17 @@ func (v *View) processMsg(sender uint64, m *protos.Message) {
        }
 }
 
+func (v *View) maybeInitializeFutureMsgVoteSetForSeq(msgProposalSeq uint64) {
+       if _, exists := v.futureMsgs[msgProposalSeq]; !exists {
+               v.futureMsgs[msgProposalSeq] = &voteSet{
+                       validVote: func(_ uint64, _ *protos.Message) bool {
+                               return true
+                       },
+               }
+               v.futureMsgs[msgProposalSeq].clear(v.N)
+       }
+}
+
 func (v *View) run() {
        defer v.viewEnded.Done()
        defer func() {
@@ -247,6 +266,10 @@ func (v *View) run() {
                })
        }()
        for {
+               if m := v.receivedMsgForNextSeq(); m != nil {
+                       v.processMsg(m.sender, m.Message)
+                       continue
+               }
                select {
                case <-v.abortChan:
                        return
@@ -258,6 +281,38 @@ func (v *View) run() {
        }
 }
 
+func (v *View) receivedMsgForNextSeq() *incMsg {
+       if len(v.futureMsgs) == 0 {
+               return nil
+       }
+
+       // Perform garbage collection on stale vote sets
+       for seq := range v.futureMsgs {
+               if seq < v.ProposalSequence {
+                       delete(v.futureMsgs, seq)
+               }
+       }
+
+       // This sequence or next sequence
+       for _, seq := range []uint64{v.ProposalSequence, v.ProposalSequence+1} {
+               vs, exists := v.futureMsgs[seq]
+               if ! exists {
+                       continue
+               }
+               // No votes in this vote set
+               if len(vs.votes) == 0 {
+                       continue
+               }
+               // Take a vote out and return it
+               m := <- vs.votes
+               return &incMsg{
+                       sender: m.sender,
+                       Message: m.Message,
+               }
+       }
+       return nil
+}
+
 func (v *View) doPhase() {
        switch v.Phase {
        case PROPOSED:

```

internal/bft/view.go
```diff
@@ -258,6 +266,48 @@ func (v *View) run() {
...
 }
 
+func (v *View) maybeAddFutureMsgForSeq(msgProposalSeq uint64, sender uint64, m *protos.Message) {
+	if msgProposalSeq <= v.ProposalSequence {
+		return
+	}
+
+	fv, ok := v.futureMsgs[msgProposalSeq]
+	if !ok {
+		fv = make([]*vote, 0, 1)
```
yacovm: There is no need to "make" a slice. A nil slice is also a valid slice. The code will also work if you remove lines 274-277

internal/bft/view.go
```diff
@@ -93,6 +93,7 @@ type View struct {
 	nextPrePrepare chan *protos.Message
 	nextPrepares   *voteSet
 	nextCommits    *voteSet
+	futureMsgs     map[uint64][]*vote
```
yacovm: This []*vote approach doesn't protect you from a malicious node sending you messages over and over again and filling up your memory and spending lots of time in the for loop in line 253. 

You need to use a vote set, in fact you need to use several vote sets (one for each type of message - pre-prepare, prepare and commit) similar to what I did. 

and this current solution also doesn't protect you from having infinite amount of sequences.

internal/bft/view.go
pfi79: Thank you. I had forgotten.

internal/bft/view.go
pfi79: I'll redo it, but in a slightly different way.

pfi79: @yacovm PrePrepare can only be from the leader?

yacovm: Yes

internal/bft/view.go
```diff
@@ -258,6 +266,77 @@ func (v *View) run() {
...
+		vf = &futureMsgsSet{
+			isPrePrepsre: false,
+			prepares:     make(map[uint64]struct{}, v.N),
+			commits:      make(map[uint64]struct{}, v.N),
+			msgs:         make([]*vote, 0, v.N*2+1),
+		}
+		v.futureMsgs[msgProposalSeq] = vf
+	}
+
+	if pp := m.GetPrePrepare(); pp != nil && vf.isPrePrepsre {
```
yacovm: why not do something similar to what you with prepares and commits with the pre-prepares?

internal/bft/view.go
```diff
@@ -258,6 +266,77 @@ func (v *View) run() {
 	}
 }
 
+func (v *View) maybeAddFutureMsgForSeq(msgProposalSeq uint64, sender uint64, m *protos.Message) {
+	if msgProposalSeq <= v.ProposalSequence {
```
yacovm: what if you get a message for a sequence that is 1000000 higher than your current sequence?

internal/bft/view.go
pfi79: I don't know which of the prePrepare then give to the processing. Because I don't know who the leader of the n+1 epoch is. Is there a check for this? What happens if more than 1 prePrepare message arrives?
If my doubts are resolved, I'll redo it.

internal/bft/view.go
pfi79: ok. I agree with the comment. I'll redo it.

internal/bft/view.go
yacovm: The leader of the n+1 epoch will be a leader of a different view, not this view. 
When the leader changes, the view is destroyed and a new view is built.

And it doesn't really matter in any case, because you record who sent it and you invoke **processPrePrepare** it checks if the node that sent it, is really the leader.

pfi79: I looked up the code. There can only be one prePrepare in an era. I have not found where extra prePrepare is cut off.

pfi79: So in this view the leader of any era is the same? Exactly. I didn't know, and I also forgot.

yacovm: Each sequence has a single pre-prepare, but what I meant is that when a leader is rotated, the view is terminated and a new one is started.

pfi79: I tweaked this code

internal/bft/view.go
```diff
@@ -258,6 +266,81 @@ func (v *View) run() {
...
+			isPrePrepsre: false,
+			prepares:     make(map[uint64]struct{}, v.N),
+			commits:      make(map[uint64]struct{}, v.N),
+			msgs:         make([]*vote, 0, v.N*2+1),
+		}
+		v.futureMsgs[msgProposalSeq] = vf
+	}
+
+	if pp := m.GetPrePrepare(); pp != nil {
+		if sender != v.LeaderID || vf.isPrePrepsre {
```
yacovm: This protection (*if sender != v.LeaderID || vf.isPrePrepsre*) is not needed because any byzantine node can theoretically send a pre-prepare or send multiple pre-prepares. 

We protect against such behavior in the rest of the code. All this new code needs to do, is to correctly save the messages and then invoke **processMsg** so if we handle such cases in the rest of the code then we don't need to add extra checks here.

An equivalent and similar argument is that if this check is **needed** here, then there is some problem in the rest of the code (which I don't believe is the case).

internal/bft/view.go
```diff
@@ -258,6 +266,81 @@ func (v *View) run() {
...
+			return
+		}
+
+		vf.isPrePrepsre = true
```
yacovm: typo in **isPrePrepsre** and in any case we don't need it because once the view has processed a pre-prepare it will never process it again.

internal/bft/view.go
pfi79: I don't quite understand.
Should we leave it like this?
```
if vf.isPrePrepsre {
	return
}
vf.isPrePrepsre = true
```

You have one channel for the prePrepare equal to capacity everywhere, you can't write the second one there. They will be eliminated. I've sifted them here, too, so as not to store unnecessary things.

yacovm: I think that you can handle the pre-prepares the same way you handle the prepares and the commits. 
You don't need to care if you get it from the leader or not in this code you're adding because the rest of the code already as the required checks.

pfi79: ок

internal/bft/view.go
```diff
@@ -258,6 +266,82 @@ func (v *View) run() {
 	}
 }
 
+func (v *View) maybeAddFutureMsgForSeq(msgProposalSeq uint64, sender uint64, m *protos.Message) {
+	if msgProposalSeq <= v.ProposalSequence || msgProposalSeq > v.ProposalSequence+10 {
```
C0rWin: What is this `10` here? Is this to limit how many sequences forward would keep? Please do it with contrast and give a meaningful name.

test/basic_test.go
```diff
@@ -1641,7 +1639,9 @@ func TestBlacklistAndRedemption(t *testing.T) {
 
 	// Ensure we remain on view 1, and that node 1 is in the blacklist
 	md := &smartbftprotos.ViewMetadata{}
+	nodes[1].ldLock.RLock()
```
yacovm, line 1642: why did you need to lock this here? Do you get a data race without?

internal/bft/util.go
```diff
@@ -141,6 +141,13 @@ func (vs *voteSet) registerVote(voter uint64, message *protos.Message) {
 	vs.votes <- &vote{Message: message, sender: voter}
 }
 
+type futureMsgsSet struct {
+	prePrepsres map[uint64]struct{}
```
yacovm: still a type (prePrepsres)

test/basic_test.go
pfi79: Yes. This is where the data race comes in.

yacovm:
Code looks OK to me but I would want to examine at it more cautiously next week, namely pull it locally and take a closer look

yacovm:
Sorry, I'll check it early next year. This week I was busy with https://github.com/SmartBFT-Go/consensus/pull/481 and https://github.com/SmartBFT-Go/consensus/pull/484 so didn't have time to look.

test/basic_test.go
```diff
@@ -1761,7 +1766,9 @@ func TestBlacklistMultipleViewChanges(t *testing.T) {
 
 	// Ensure we remain on view 2, and that nodes 2,3 are in the blacklist
 	md := &smartbftprotos.ViewMetadata{}
+	nodes[5].ldLock.RLock()
```
yacovm, line 1769: I guess this wasn't caught by the race detector because we're running Go 1.16 in CI. I don't know with what Go version you ran it, but even with Go 1.17.5 it isn't caught in the race detector.
Regardless, this isn't related to the change and is a separate topic.

While this indeed may solve a potential data race, this isn't the right solution.

The problem is that the test is written wrong. In line 1763 we receive the *AppRecord from the delivered channel, and we can check the metadata by looking at it. All that is required is to save the content that we read into a variable:

```
   latestDecisions := make(map[uint64]*AppRecord)
   ...
   latestDecisions[i] := <-nodes[i].Delivered
```

Same applies for line 1801.

pfi79: I've had it show up three times out of 50 times.
I don't know what the test does. 
Just decided to fix it quickly.

internal/bft/view.go
```diff
@@ -202,6 +206,7 @@ func (v *View) processMsg(sender uint64, m *protos.Message) {
 	// This message is either for this proposal or the next one (we might be behind the rest)
 	if msgProposalSeq != v.ProposalSequence && msgProposalSeq != v.ProposalSequence+1 {
 		v.Logger.Warnf("%d got message from %d with sequence %d but our sequence is %d", v.SelfID, sender, msgProposalSeq, v.ProposalSequence)
+		v.maybeAddFutureMsgForSeq(msgProposalSeq, sender, m)
 		v.discoverIfSyncNeeded(sender, m)
```
yacovm, line 210: Notice that in **discoverIfSyncNeeded** we try to detect if there exists a threshold of commits on a message with a higher sequence and if that's the case then we stop the view and sync.

Given this fact, what's the point of saving messages up to 10 sequences ahead and then re-inserting them, if this view is likely to be terminated anyway once it detects a commit of a future block? 

Either way (whether I am wrong or right) I think we need an integration test (in **test/basic_test.go**) that shows how the system behaves in such a situation in which it receives messages for future sequences.

test/basic_test.go
yacovm: No problem, then leave it out and either you or we will take care of it in a separate PR.

internal/bft/view.go
pfi79: 10 is the margin.
Now the tests are being conducted and the system starts to shake at high loads. How it will behave in the future, neither you nor I know. I re-balanced for the case that the node normally receives messages only from f nodes (well, such uneven and retarded bandwidth communication lines)

test/basic_test.go
pfi79: Because of them I do not always pass PR tests. And I can not restart them. And it makes me very nervous.

yacovm: All right then can you please just read the AppRecord from the delivered channel like in the example above? While adding a lock like you did works, it does not address the fundamental problem that exists in the test and is the reason for the data race.

internal/bft/view.go
yacovm: What I am trying to say is that I think that a different part of the code that is not touched by your PR (**discoverIfSyncNeeded**) will make the changes in your PR non-existent because upon detection of a commit of a future block, it stops the view and syncs. 

> How it will behave in the future, neither you nor I know.

We should discover how the change in your PR affects the system by adding an integration test to **test/basic_test.go** which checks the scenario you are addressing. With not just three consecutive proposals but more, at least 5 or 6.

test/basic_test.go
pfi79: OK, I will.

internal/bft/view.go
pfi79: I will add 5 or 6 to the test.
We'll discuss afterwards.

test/basic_test.go
pfi79: I checked the code. The problem seems to be that the variable is changed in one goroutine, but can be read from other goroutines. Isn't it? test/test_app.go:264

yacovm: Right and we read it in the test. We don't need to read it in that way in the test, because we anyway read it from the channel when we do **<-nodes[i].Delivered** (and then it's memory coherent). So we just need to save what we read from the channel into a variable.

PR #483, opened 2021-12-29
[WIP] Add BFT metrics

muzykantov:


pkg/api/diagnostics.go
```diff
@@ -0,0 +1,227 @@
...
+	// With is used to provide label values when recording a Histogram
+	// observation. This must be used to provide values for all LabelNames
+	// provided to HistogramOpts.
+	With(labelValues ...string) Histogram
+	Observe(value float64)
+}
+
+// Diagnostics contains set of various metrics and logging interfaces
+type Diagnostics struct {
+	l Logger
```
yacovm, line 47: Why is the logger part of the diagnostics?

internal/bft/requestpool.go
```diff
@@ -260,7 +273,7 @@ func (rp *Pool) NextRequests(maxCount int, maxSizeBytes uint64, check bool) (bat
 	fullC := len(batch) == maxCount
 	full = fullS || fullC
 	if len(batch) > 0 {
-		rp.logger.Debugf("Returning batch of %d requests totalling %dB",
+		rp.diag.L().Debugf("Returning batch of %d requests totalling %dB",
```
yacovm, line 276: why do logging events pass through metrics?

pkg/api/diagnostics.go
C0rWin: This will be changed and yes, no reason to have logger within diagnostics.

internal/bft/requestpool.go
C0rWin: it should be refactored.

C0rWin (changes requested):
There is a need to refactor the change to extract logger from Diagnostics and also create abstractions to enable injections of metrics providers into lib

PR #484, opened 2021-12-30
Synchronize Sync and view change proposal deliver to prevent a double commit

yacovm:
This commit fixes a bug in which a node is unaware of a proposal being committed
by the rest of the nodes, and it attempts a view change which fails, and triggers
a view change timeout, which in turn triggers a sync.

the asynchronous sync procedure may then commit that proposal, and the proposal
will later on be attempted to be committed again once the view change resumes
its operation, because the two procedures are parallel.

The easy and trivial way of solving this would be to make the view change
block until the sync process has ended. While this might work, it makes
the single-threaded view change be dependant on network I/O, which is
might slow down view change for other nodes due to votes being late.

Instead, this commit highjacks the view change commit process and:
1) Waits for any active sync to finish and prevents any syncs from occuring
2) Checks the latest sequence committed and if the proposal being attempted
   to be committed by the view change is stale, it calls sync which
   would return either the last proposal committed, or a later one.

Signed-off-by: Yacov Manevich <yacovm@il.ibm.com>


internal/bft/viewchanger.go
```diff
@@ -1126,6 +1130,20 @@ func (v *ViewChanger) deliverDecision(proposal types.Proposal, signatures []type
 	v.Pruner.MaybePruneRevokedRequests()
 }
 
+func (v *ViewChanger) isProposalLatestComparedToCheckpoint(proposal types.Proposal) bool {
+	checkpointProposal, _ := v.Checkpoint.Get()
+	return v.sequenceFromProposal(proposal.Metadata) >= v.sequenceFromProposal(checkpointProposal.Metadata)
```
C0rWin: shouldn't it be strictly greater than?

yacovm: Yes, better that way.

C0rWin (approved):

PR #531, opened 2023-04-23
add metrics for pool request

pfi79:


internal/bft/requestpool.go
```diff
@@ -72,6 +73,7 @@ type Pool struct {
 type requestItem struct {
 	request []byte
 	timeout *time.Timer
+	begin   time.Time
```
C0rWin: Can you give a more descriptive name? 
```suggestion
	additionTimestamp   time.Time
```

pfi79: fix

C0rWin (approved):

PR #551, opened 2023-06-04
delete mock ViewChanged

pfi79:


internal/bft/viewchanger_test.go
```diff
@@ -2082,14 +2082,7 @@ func TestCommitWhileHavingInFlight(t *testing.T) {
...
 	verifier.On("RequestsFromProposal", mock.Anything).Return(nil)
 	controller := &mocks.ViewController{}
-	viewNumChan := make(chan uint64)
-	seqNumChan := make(chan uint64)
-	controller.On("ViewChanged", mock.Anything, mock.Anything).Run(func(args mock.Arguments) {
-		num := args.Get(0).(uint64)
-		viewNumChan <- num
-		num = args.Get(1).(uint64)
-		seqNumChan <- num
-	}).Return(nil).Once()
```
C0rWin, line 2092: Here I think the problem is that test is missing assertion on the values of the view change, hence need to add checks to read and validate view number and sequence instead of deleting this code.

C0rWin (approved):

PR #552, opened 2023-06-04
waiting for a change of leaders

pfi79:


test/basic_test.go
```diff
@@ -425,7 +425,7 @@ func TestMultiLeadersPartition(t *testing.T) {
 	}
 
 	for i := 2; i < numberOfNodes; i++ {
-		assert.GreaterOrEqual(t, uint64(3), nodes[i].Consensus.GetLeaderID())
+		assert.LessOrEqual(t, uint64(3), nodes[i].Consensus.GetLeaderID())
```
C0rWin: I do not understand this change, how come it could be less? In the test node ID = 1 and node ID = 2 are both partitioned, hence could not be a leader, making here a change to `LessOrEqual` you actually changing assertion to make test to pass even though not all nodes switched to a new leader, which is a mistake.

test/basic_test.go
```diff
@@ -240,7 +240,7 @@ func TestLeaderInPartition(t *testing.T) {
 	for i := 0; i < numberOfNodes-2; i++ {
 		assert.Equal(t, data[i], data[i+1])
 	}
-	assert.Equal(t, uint64(2), nodes[2].Consensus.GetLeaderID())
+	assert.LessOrEqual(t, uint64(2), nodes[2].Consensus.GetLeaderID())
```
C0rWin, line 243: If leaderID here is less than 2 it means node still thinks leader is node with ID = 1, which is error or bug, making here a change to `LessOrEqual` just hides this bug/error.

test/basic_test.go
pfi79: 1. description of function LessOrEqual from source:
```
// LessOrEqual asserts that the first element is less than or equal to the second
//
//    assert.LessOrEqual(t, 1, 2)
//    assert.LessOrEqual(t, 2, 2)
//    assert.LessOrEqual(t, "a", "b")
//    assert.LessOrEqual(t, "b", "b")
```
2. In our case, I expect a positive outcome if the leader is all 3 or all 4
3. Why isn't it strictly 3? Because the `func accelerateTime` works. Nodes can timeout to slip a state where the leader should become the 3rd node and become the 4th node.

test/basic_test.go
pfi79: In our case, I expect a positive outcome if the leader is all 2 or 3 or 4

C0rWin: OK, so how do you know all reached same number?

test/basic_test.go
C0rWin: How do you know all decided about same leader?

test/basic_test.go
pfi79: in this case, there is no loop.

test/basic_test.go
pfi79: fix

C0rWin (approved):

PR #553, opened 2023-06-04
fix committed sequence twice

pfi79:
please check these changes very carefully

### DIFF
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -101,6 +101,7 @@ type Controller struct {
 	Logger             api.Logger
 	Assembler          api.Assembler
 	Application        api.Application
+	Deliver            api.Application
 	FailureDetector    FailureDetector
 	Synchronizer       api.Synchronizer
 	Signer             api.Signer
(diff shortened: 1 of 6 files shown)


internal/bft/controller.go
```diff
@@ -535,12 +536,11 @@ func (c *Controller) run() {
 func (c *Controller) decide(d decision) {
 	c.Logger.Debugf("Delivering to app from Controller decide the last decision proposal")
 	begin := time.Now()
-	reconfig := c.Application.Deliver(d.proposal, d.signatures)
+	reconfig := c.Deliver.Deliver(d.proposal, d.signatures)
 	c.MetricsView.LatencyBatchSave.Observe(time.Since(begin).Seconds())
```
HagarMeir: Maybe move the metrics to be only on the application deliver

internal/bft/controller.go
```diff
@@ -954,7 +954,7 @@ func (med *MutuallyExclusiveDeliver) Deliver(proposal types.Proposal, signature
 	// do not proceed to commit the proposal, but instead invoke a sync and update the checkpoint once more
 	// to match the sync result.
 	latest := med.C.latestSeq()
-	if latest >= pendingProposalMetadata.LatestSequence {
+	if latest != 0 && latest >= pendingProposalMetadata.LatestSequence {
```
HagarMeir, line 955: Why do you check if `latest != 0`

internal/bft/controller.go
pfi79: Yes you're right, I'll correct that.

If you don't mind, I'll correct this one in another pr

internal/bft/controller.go
pfi79: When latest == 0 it is both "sequence is 0" and "no sequence". 
I did this to go from "no sequence" to "first sequence with index 0" state.

internal/bft/controller.go
pfi79: I corrected that.

C0rWin (approved):

PR #556, opened 2023-06-29
controller deliver and set checkpoint under a lock

HagarMeir:


pfi79:
not all Checkpoint.Set calls are mutex-covered
- https://github.com/SmartBFT-Go/consensus/blob/d6a0ebcc28f8a305da94483daa1421b59de90328/internal/bft/viewchanger.go#L1178
- https://github.com/SmartBFT-Go/consensus/blob/d6a0ebcc28f8a305da94483daa1421b59de90328/internal/bft/viewchanger.go#L1335

PR #560, opened 2023-07-17
fix two twice delivery on sync

pfi79:
Proposal delivery can come from 2 goroutines: Controller and ViewChanger. 
4 places:
- https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/controller.go#L537
- https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/controller.go#L594
- https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/viewchanger.go#L1171
- https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/viewchanger.go#L1311

Goroutines are fairly independent. It is required to ensure that delivery is not at the same time and that there is no twice delivery.
To prevent at the same time delivery, all deliveries are performed from under the [syncLock](https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/controller.go#L143) mutex.
To prevent twice delivery, the [Checkpoint](https://github.com/SmartBFT-Go/consensus/blob/v2/internal/bft/controller.go#L110) parameter is implemented. Before delivery, the code reads Checkpoint and decides if the proposal should be delivered. After delivery, the code must set a new value in Checkpoint. Without this, double delivery is possible.

Found the last place where proposals can be delivered and Checkpoint is not set. This pull request fixes this error.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   pull_request:
-    
+
 jobs:
 
   build:
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -630,6 +630,10 @@ func (c *Controller) sync() (viewNum uint64, seq uint64, decisions uint64) {
 
 	latestSequence := c.latestSeq()
 
+	c.Logger.Debugf("Node %d is setting the checkpoint after sync to view %d and seq %d", c.ID, md.ViewId, md.LatestSequence)
+	c.Checkpoint.Set(decision.Proposal, decision.Signatures)
+	c.verificationSequence = uint64(decision.Proposal.VerificationSequence)
+
 	if md.ViewId < c.currViewNumber {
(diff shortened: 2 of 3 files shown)


C0rWin:
Can you please provide a unit-test which clearly demonstrates the issue you are addressing in the commit?

pfi79:
> Can you please provide a unit-test which clearly demonstrates the issue you are addressing in the commit?

pfi79:
> Can you please provide a unit-test which clearly demonstrates the issue you are addressing in the commit?

So far I have not been able to create a test in which this error always appears.
But I will add the [test.log](https://github.com/SmartBFT-Go/consensus/files/12267971/test.log) of the test in which this error occurred

pfi79:
> Can you please provide a unit-test which clearly demonstrates the issue you are addressing in the commit?

I added a test.

C0rWin (approved):
Looks good to me

PR #568, opened 2023-11-26
fix 2 test - lose commits message

pfi79:
Node operation is interrupted after processing prepare messages. 
But we focus on logs. When commit messages may have already arrived.
I added the part that commit messages will definitely not arrive.

### DIFF
--- a/test/basic_test.go
+++ b/test/basic_test.go
@@ -1272,6 +1272,10 @@ func TestLeaderCatchUpWithoutSync(t *testing.T) {
 	restoredWG := sync.WaitGroup{}
 	restoredWG.Add(1)
 
+	var blockCommits uint32
+	nodes[0].LoseMessages(func(msg *smartbftprotos.Message) bool {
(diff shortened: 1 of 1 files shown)


C0rWin:
> But we focus on logs. When commit messages may have already arrived.
> I added the part that commit messages will definitely not arrive.

Can you please explain a little bit more? What is the problem with commit message here?

pfi79:
> > But we focus on logs. When commit messages may have already arrived.
> > I added the part that commit messages will definitely not arrive.
> 
> Can you please explain a little bit more? What is the problem with commit message here?

The author of the test clearly had a desire to interrupt the node after processing the prepare messages. To check how the node will restore its work (e.g. synchronise) after restart. 

For this purpose, the author intercepted the message about processing of prepare messages in the logger of the goroutine view. 

But it happens that commit messages are already in the controller's goroutine. And the node manages to process them before completion.
And after restarting the node, the expected situation never occurs.

I fixed this situation by prohibiting the node from receiving commit messages until a certain moment

C0rWin (approved):

PR #569, opened 2023-11-26
Sometimes there isn't enough time to wait for an event.

pfi79:
Increased the time.

Fixed working with break operator

### DIFF
--- a/test/basic_test.go
+++ b/test/basic_test.go
@@ -2663,7 +2663,7 @@ func TestViewChangeAfterTryingToFork(t *testing.T) {
 	nodes[4].Connect()
 
 	// Waiting for a real change of leader and view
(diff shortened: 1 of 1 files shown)


test/basic_test.go
```diff
@@ -2663,7 +2663,7 @@ func TestViewChangeAfterTryingToFork(t *testing.T) {
 	nodes[4].Connect()
 
 	// Waiting for a real change of leader and view
-	fail = time.After(1 * time.Minute)
+	fail = time.After(90 * time.Second)
```
C0rWin: please make it to variable to not change next time in three places.

pfi79: fix

C0rWin (approved):

PR #570, opened 2024-01-17
Deliver Twice Bug Fix

pfi79:
In this scenario a decision is delivered twice.
Once by the synchronizer and a second time by the view data received during a view change.

The reason this is even possible is twofold.
First of all, after a view change timeout, we do not stop the running view (since a sync may created a good view), and so the view and the view change are running concurrently. It is feasible for a decision to be made during the concurrent view and a view change to occur while the new view message doesn't include that decision. That decision will be delivered to a node during a sync. Secondly, when the sync returns with the decision belonging to an old view we do not update the checkpoint. And so when there is another view change and a view data msg with that decision is received the node thinks this is a new decision (because its checkpoint was not updated). Finally, the node delivers that decision again from the view data msg (for the second time).

The fix includes an update of the checkpoint also when the synchronizer returns with an old view, and aborting a view before sending a view data msg to avoid running concurrent view and view change.


C0rWin (dismissed):

C0rWin:
@pfi79, can you please point to the original commit as a reference in the commit message?

pfi79:
> @pfi79, can you please point to the original commit as a reference in the commit message?

https://github.com/fabric-research/consensus/commit/a126c0e8ac537d6bd53c29a9d7a2737c1670e78c

pfi79:
> @pfi79, can you please point to the original commit as a reference in the commit message?

I added

C0rWin (approved):

PR #571, opened 2024-02-09
Change options of the pool after reconfig

pfi79:


C0rWin (commented):
It would be nice if you can add unit test to demonstrate the change is working as expected.

pfi79:
> It would be nice if you can add unit test to demonstrate the change is working as expected.

I added

C0rWin (approved):

PR #599, opened 2024-10-11
fix error prevs commit

pfi79:


yacovm (changes requested):
1. Can you add a commit message? 
2. Can you please add a test that checks that when we send pre-prepare, we don't only attach the previous commit signatures when periodical leader rotation is enabled?

pfi79:
> 1. Can you add a commit message?
> 2. Can you please add a test that checks that when we send pre-prepare, we don't only attach the previous commit signatures when periodical leader rotation is enabled?

done

internal/bft/view_test.go
```diff
@@ -145,6 +145,98 @@ func TestViewBasic(t *testing.T) {
...
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			basicLog, err := zap.NewDevelopment()
+			assert.NoError(t, err)
+			log := basicLog.Sugar()
+			comm := &mocks.CommMock{}
+			commWG := sync.WaitGroup{}
+			comm.On("BroadcastConsensus", mock.Anything).Run(func(args mock.Arguments) {
+				fmt.Println("Sending", args.Get(0))
```
yacovm: please remove the print

pfi79: Ugh, my inattention.

Fixed it.

yacovm (approved):

PR #655, opened 2025-11-24
fix select version protoc for install

pfi79:


.github/workflows/build.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 env:
   GO_VER: 1.25.3
-  PROTOC_VER: 33.0
+  PROTOC_VER: "33.0"
```
tock-ibm, line 12: @pfi79  why is it a problem? is in interpreted as a number instead of a version?

pfi79: the [robot tried to throw a pr](https://github.com/hyperledger-labs/SmartBFT/pull/653/checks) that failed validation. The installation of protocol 33.0 is specified, but action installed 33.1. For some reason they missed a significant 0. The quotation marks made it so that action would not skip this zero.

here is an example of a version change error
image

pfi79: I'm trying to fix this situation, and then I'll restart the robot pipes.

tock-ibm (approved):

PR #663, opened 2026-02-11
Fix DecisionsInView reset to zero during same-height sync

ivan-atme:
When sync() is called but the synchronizer returns the same block height as the controller ("already at target height"), newDecisionsInView stays at its zero-initialized value because the update at line 641 is guarded by `latestDecisionSeq > controllerSequence`, which is false when equal.

This zero is passed to changeView(), resetting DecisionsInView from its correct value (e.g. 9578) to 0. The next proposal from the leader carries the correct DecisionsInView, fails validation at view.go:577, and is rejected as "bad proposal", triggering a recovery sync loop that costs ~10 seconds per cycle and can cause Hyperledger Fabric orderers to fall permanently behind.

Fix: decouple the DecisionsInView update from the checkpoint update by moving it to a separate `if latestDecisionMetadata != nil` block, so decisions are always derived from sync response metadata regardless of whether the sequence advanced.

### DIFF
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -632,14 +632,14 @@ func (c *Controller) sync() (viewNum uint64, seq uint64, decisions uint64) {
 	controllerViewNum := c.currViewNumber
 	newViewNum = controllerViewNum
 
+	newDecisionsInView = c.getCurrentDecisionsInView()
+
 	if latestDecisionSeq > controllerSequence {
 		c.Logger.Infof("Synchronizer returned with sequence %d while the controller is at sequence %d", latestDecisionSeq, controllerSequence)
 		c.Logger.Debugf("Node %d is setting the checkpoint after sync returned with view %d and seq %d", c.ID, latestDecisionViewNum, latestDecisionSeq)
 		c.Checkpoint.Set(latestDecision.Proposal, latestDecision.Signatures)
 		c.verificationSequence.Store(uint64(latestDecision.Proposal.VerificationSequence))
 		newProposalSequence = latestDecisionSeq + 1
-	}
-	if latestDecisionMetadata != nil {
 		newDecisionsInView = latestDecisionDecisions + 1
 	}
 
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -638,6 +638,8 @@ func (c *Controller) sync() (viewNum uint64, seq uint64, decisions uint64) {
 		c.Checkpoint.Set(latestDecision.Proposal, latestDecision.Signatures)
 		c.verificationSequence.Store(uint64(latestDecision.Proposal.VerificationSequence))
 		newProposalSequence = latestDecisionSeq + 1
+	}
+	if latestDecisionMetadata != nil {
 		newDecisionsInView = latestDecisionDecisions + 1
 	}
 
--- a/internal/bft/controller_sync_test.go
+++ b/internal/bft/controller_sync_test.go
@@ -30,20 +30,13 @@ func (s *stubComm) SendTransaction(targetID uint64, request []byte)  {}
 func (s *stubComm) Nodes() []uint64                                  { return []uint64{1, 2, 3, 4} }
 func (s *stubComm) BroadcastConsensus(m *protos.Message)             {}
 
-// TestSyncDecisionsInView tests that the sync() method correctly preserves
-// the DecisionsInView counter. This test reproduces a bug where sync()
-// resets DecisionsInView to 0 when the synchronizer returns the same block
-// height as the controller (the "already at target height" case).
-//
-// Bug location: controller.go, sync() function.
-// Line 610: newDecisionsInView is zero-initialized.
-// Line 635: only updated when latestDecisionSeq > controllerSequence.
-// When latestDecisionSeq == controllerSequence, newDecisionsInView stays 0.
-//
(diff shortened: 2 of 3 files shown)


HagarMeir:
Thank you for discovering this bug and providing a fix.
Can you please attach a log with the events you described?

ivan-atme:
@HagarMeir I am attaching orderers' logs: [org0-orderer-1-stuck.zip](https://github.com/user-attachments/files/25361999/org0-orderer-1-stuck.zip) .  

Sorry, logs are quite long (05:59:38-06:02:31). There is a reason for this (see below). 

Additional context about the problem: 

- New blocks are stable arriving with approximate speed 15 blocks/min. 
- The problem recurring weekly when one of orderers suddenly gets stuck.
- Stuck orderer restart returns it to be in sync. 
- The problem in these logs I am going to solve is that org0-orderer-1 got stuck in "otf" channel at height 2365915.

### Timeline

| Time | Event |
|------|-------|
| 05:59:39–05:59:50 |  The bug I fixed currently happened here: DecisionsInView=0, proposal rejected, recovery sync |
| 05:59:50-06:00:01 | The bug reproduced again, but with faster recovery sync |
| | Initially I was going to cut logs here |
| 06:00:45-06:01:45 | But today I noticed the fact that not one but all orderers temporary got stuck at height 2365915 here |
| 06:01:45–06:02:30 | Other orderers started producing new blocks, and only org0-orderer-1 is still stuck on 2365915 height (until the end of logs) |

### Conclusion
Thus, currently fixed issue was not the only cause which triggered the org0-orderer-1  got stuck. I found few more issues that together make orderer being stuck. I described two more changes I offer in RFC draft: https://gist.github.com/ivan-atme/73c430c9a530985546af6f34139cddb6#file-rfc-smartbft-controller-bugs-upstream-md

Please let me know if there's anything I can do to make the review easier.
Thanks for your time!

HagarMeir:
maybe you can just copy the part of the log showing this bug to here? 
this part shouldn't be long

pfi79:
I will support @HagarMeir . I don't quite understand what's being fixed here. And I don't trust artificial intelligence. I'm still using the natural one.

internal/bft/controller.go
```diff
@@ -638,6 +638,8 @@ func (c *Controller) sync() (viewNum uint64, seq uint64, decisions uint64) {
 		c.Checkpoint.Set(latestDecision.Proposal, latestDecision.Signatures)
 		c.verificationSequence.Store(uint64(latestDecision.Proposal.VerificationSequence))
 		newProposalSequence = latestDecisionSeq + 1
+	}
+	if latestDecisionMetadata != nil {
```
HagarMeir: a better fix would be to set `newDecisionsInView` to `controllerDecisionsInView+1` just like we do with the sequence in line 630

internal/bft/controller_sync_test.go
```diff
@@ -0,0 +1,159 @@
...
+// Bug location: controller.go, sync() function.
+// Line 610: newDecisionsInView is zero-initialized.
+// Line 635: only updated when latestDecisionSeq > controllerSequence.
+// When latestDecisionSeq == controllerSequence, newDecisionsInView stays 0.
+//
+// Real-world impact: after sync returns same height, changeView() resets
+// DecisionsInView from ~9578 to 0. The next proposal carries the correct
+// DecisionsInView=9578, which fails validation at view.go:577, causing the
+// orderer to reject valid proposals and enter a recovery sync loop.
+func TestSyncDecisionsInView(t *testing.T) {
```
HagarMeir, line 40: please also add a test in `test/basic_test.go`

HagarMeir: this test should fail without the fix you are offering (and should pass with it)

ivan-atme:
Sorry for the answer delay. Here is the exact line in logs revealing current PR bug (Sync returns same height, DecisionsInView resets to 0):

```
2026-02-11T05:59:50.167415422Z {"level":"warn","name":",fabric-orderer-test3-org0-orderer-1-0","caller":"bft/view.go:578","msg":"Expected decisions in view 0 but got 9578","channel":"otf"}
2026-02-11T05:59:50.167421922Z {"level":"warn","name":",fabric-orderer-test3-org0-orderer-1-0","caller":"bft/view.go:388","msg":"1 received bad proposal from 6: invalid decisions in view","channel":"otf"}
```

In my current changes instead of `checkpoint.DecisionsInView + 1 `(mirroring the sequence pattern), I used `c.getCurrentDecisionsInView()` which reads the controller's in-memory value directly.

The reason is that the +1 derivation from checkpoint doesn't hold on fresh systems. After decide(), the checkpoint stores DecisionsInView = N (the value in the proposal metadata, before the increment), while currDecisionsInView is already N + 1. So checkpoint + 1 works correctly after at least one commit. But when no commits have happened yet (checkpoint DecisionsInView = 0, currDecisionsInView = 0), the +1 approach gives 1 instead of 0. This mismatch causes changeView() to spuriously restart the view instead of detecting it as already running.

I verified this with `TestTryCommittedSequenceTwice` — it fails (hangs) with the checkpoint + 1 approach due to this exact edge case, but passes with getCurrentDecisionsInView().

Also added the integration test `TestSyncSameHeightPreservesDecisionsInView` in `test/basic_test.go` as requested — it fails without the fix and passes with it.

HagarMeir (approved):

PR #670, opened 2026-05-01
add v2 for version  v2

pfi79:


tock-ibm:
@pfi79 @C0rWin Don't you prefer to just retag v2.3.* to v1.*  (and delete v2.3.*)? it would save everybody from changing all the import paths in fabric, fabric-x and every other project using smartbft... ? 
No one is explicitly dependent on `@v2.3.*` now any way so it won't make a difference to anyone, I think.

PR #681, opened 2026-08-09
Avoid blocking on completion signals when the waiter has left

kevindharmawan:
The controller and the view changer both used unbuffered channels to report completion, where the receiver can legitimately stop waiting and leave the sender blocked forever.

Controller: `Decide` stops waiting on `deliverChan` when the view aborts, but `decide` was still blocked sending on it, so the run loop stalled until shutdown. Now uses a per-decision `delivered` channel closed after delivery. In the aborted case `decide` no longer returns early, so the decision-count increment, rotation check, pruning and leader-token acquisition now run, since the decision was in fact delivered.

View changer: once `commitInFlightProposal` returned via timeout or stop, nothing read `inFlightDecideChan` or `inFlightSyncChan`. A late `Decide` or `Sync` then blocked the in-flight view's run goroutine, which blocked `Abort()` since it waits on that goroutine, which blocked the view changer in its own cleanup. No later attempt could start to drain the channel. `Decide` could at least escape on `stopChan`, while `Sync` had no escape at all. Now each attempt owns an `inFlightAttempt` holding its own buffered `decideCh` and `syncCh`, cleared when the attempt ends, with non-blocking sends. The in-flight view's `Decider` and `Sync` are callbacks bound to that attempt at creation, so a late `Decide` or `Sync` can only signal its own attempt and cannot disturb a later one.

Adds regression tests for all three completion paths.

### DIFF
--- a/internal/bft/controller.go
+++ b/internal/bft/controller.go
@@ -132,7 +132,6 @@ type Controller struct {
 
 	syncChan             chan struct{}
 	decisionChan         chan decision
-	deliverChan          chan struct{}
 	leaderToken          chan struct{}
 	verificationSequence atomic.Uint64
 
@@ -542,9 +541,10 @@ func (c *Controller) decide(d decision) {
 	}
 	c.Logger.Debugf("Node %d delivered proposal", c.ID)
 	c.removeDeliveredFromPool(d)
-	select {
-	case c.deliverChan <- struct{}{}:
-	case <-c.stopChan:
+	if d.delivered != nil {
+		close(d.delivered)
+	}
+	if c.stopped() {
 		return
 	}
 	c.incrementCurrentDecisionsInView()
@@ -797,7 +797,6 @@ func (c *Controller) Start(startViewNumber uint64, startProposalSequence uint64,
 	c.stopChan = make(chan struct{})
 	c.leaderToken = make(chan struct{}, 1)
 	c.decisionChan = make(chan decision, 1)
-	c.deliverChan = make(chan struct{})
 	c.viewChange = make(chan viewInfo, 1)
 	c.abortViewChan = make(chan uint64, 1)
 
@@ -882,11 +881,13 @@ func (c *Controller) stopped() bool {
 
 // Decide delivers the decision to the application
 func (c *Controller) Decide(proposal types.Proposal, signatures []types.Signature, requests []types.RequestInfo) {
+	delivered := make(chan struct{})
 	select {
 	case c.decisionChan <- decision{
 		proposal:   proposal,
 		requests:   requests,
 		signatures: signatures,
+		delivered:  delivered,
 	}:
 	case <-c.stopChan:
 		// In case we are in the middle of shutting down,
@@ -895,7 +896,7 @@ func (c *Controller) Decide(proposal types.Proposal, signatures []types.Signatur
 	}
 
 	select {
-	case <-c.deliverChan: // wait for the delivery of the decision to the application
+	case <-delivered: // wait for the delivery of the decision to the application
 	case <-c.stopChan: // If we stopped the controller, abort delivery
 	case <-c.currentViewAbortChan(): // If we stopped the view, abort delivery
 	}
@@ -918,6 +919,7 @@ type decision struct {
 	proposal   types.Proposal
 	signatures []types.Signature
 	requests   []types.RequestInfo
+	delivered  chan struct{}
 }
 
 // BroadcastConsensus broadcasts the message and informs the heartbeat monitor if necessary
--- a/internal/bft/viewchanger.go
+++ b/internal/bft/viewchanger.go
@@ -48,6 +48,28 @@ type change struct {
 	stopView bool
 }
 
+type inFlightAttempt struct {
+	id       uint64
+	view     uint64
+	sequence uint64
+	decideCh chan struct{}
+	syncCh   chan struct{}
+	viewRef  *View
+}
+
+type inFlightAttemptCallbacks struct {
+	vc      *ViewChanger
+	attempt *inFlightAttempt
+}
+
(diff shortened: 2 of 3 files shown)


internal/bft/controller.go
```diff
@@ -542,9 +541,10 @@ func (c *Controller) decide(d decision) {
 	}
 	c.Logger.Debugf("Node %d delivered proposal", c.ID)
 	c.removeDeliveredFromPool(d)
-	select {
-	case c.deliverChan <- struct{}{}:
-	case <-c.stopChan:
+	if d.delivered != nil {
+		close(d.delivered)
+	}
+	if c.stopped() {
```
HagarMeir, line 547: Behavior change worth confirming: previously `decide` blocked on the unbuffered `deliverChan` send when the waiter had left (view aborted), so the post-delivery bookkeeping below (`incrementCurrentDecisionsInView`, `checkIfRotate`/`changeView`, `MaybePruneRevokedRequests`, `acquireLeaderToken`) never ran in the aborted case. It now always runs once delivery happened. This unblocks the run loop (the real fix) and is defensible because the decision was in fact delivered, but it means a rotation `changeView` and leader-token acquisition can now fire for a view that is being aborted, racing the in-progress view change. `changeView` guards against going backwards (`latestView > newViewNumber`), which mitigates it, but please confirm the extra increment/rotate cannot momentarily double-count or start a spurious rotation view before the pending abort/view-change is processed.

internal/bft/viewchanger.go
```diff
@@ -48,6 +48,28 @@ type change struct {
 	stopView bool
 }
 
+type inFlightAttempt struct {
```
HagarMeir, line 51: Dead state: `id`, `view`, and `sequence` on `inFlightAttempt` (and `ViewChanger.inFlightAttemptSeq` that feeds `id`) are written but never read anywhere — attempt identity is established purely by pointer equality (`v.inFlightAttempt == attempt` in the cleanup defer). Only `decideCh`, `syncCh`, and `viewRef` are actually used. Consider dropping the unused fields and the `inFlightAttemptSeq` counter to reduce clutter.

internal/bft/viewchanger.go
```diff
@@ -1305,9 +1348,22 @@ func (v *ViewChanger) commitInFlightProposal(proposal *protos.Proposal) (success
...
-// Decide delivers to the application and informs the view changer after delivery
+func (v *ViewChanger) currentInFlightAttempt() *inFlightAttempt {
+	v.inFlightViewLock.RLock()
+	defer v.inFlightViewLock.RUnlock()
+	return v.inFlightAttempt
+}
+
+// Decide delivers to the application and informs the view changer after delivery.
+// It is kept for compatibility; in-flight views use per-attempt callbacks.
 func (v *ViewChanger) Decide(proposal types.Proposal, signatures []types.Signature, requests []types.RequestInfo) {
```
HagarMeir, line 1359: These public `Decide`/`Sync` methods (and `currentInFlightAttempt`) are now dead in production. In-flight views are wired with the per-attempt `callbacks` (`Decider: callbacks`, `Sync: callbacks`), and nothing else wires `ViewChanger` as a `Decider`/`Synchronizer` (the heartbeat handler is `c.controller`, and consensus.go never passes the ViewChanger as a Decider/Sync). Only the new tests call them. Beyond being dead code, they are a latent foot-gun: when invoked with a nil attempt they still call `Application.Deliver`/`Synchronizer.Sync` without stopping any view or signaling. Consider removing them rather than keeping them as a compatibility shim.

internal/bft/viewchanger.go
```diff
@@ -1322,8 +1378,13 @@ func (v *ViewChanger) Decide(proposal types.Proposal, signatures []types.Signatu
...
+	if attempt == nil {
+		return
+	}
 	select {
-	case v.inFlightDecideChan <- struct{}{}:
+	case attempt.decideCh <- struct{}{}:
+		return
+	default:
 		return
 	case <-v.stopChan:
```
HagarMeir, line 1389: Dead branch: because this `select` has a `default`, it never blocks, so the `case <-v.stopChan:` branch can never be the distinguishing choice — and all three branches just `return` anyway. The `stopChan` case is unreachable/pointless here (same in `syncInFlight`, which correctly omits it). Simplify to a plain non-blocking send:

```go
	select {
	case attempt.decideCh <- struct{}{}:
	default:
	}
```

HagarMeir:
🤖 This PR was reviewed by Claude. See the inline comments above for detailed findings.

internal/bft/viewchanger.go
kevindharmawan: Removed in 72dd0a2, along with the `inFlightAttemptSeq` counter.

internal/bft/viewchanger.go
kevindharmawan: Simplified to the plain non-blocking send in 72dd0a2, matching `syncInFlight`.

internal/bft/viewchanger.go
kevindharmawan: Removed in 72dd0a2, together with `currentInFlightAttempt`. The tests now call `decideInFlight` and `syncInFlight` directly with an explicit attempt, so the nil-attempt path no longer exists.

internal/bft/controller.go
kevindharmawan: Addressed in a167920.`decide` now records the view that produced the decision, and `decide` checks after delivery that this view is still current and not stopped. Otherwise it only prunes revoked requests and returns, skipping the increment, `checkIfRotate`/`changeView`, and `acquireLeaderToken`. The check runs on the controller run loop, the same goroutine that handles `abortView` and `changeView`, so a view change already processed is always observed first. For a dead view the result matches the pre-PR behavior, where the blocked send prevented the bookkeeping, minus the hang.

`TestControllerDecideSkipsViewBookkeepingIfDecidingViewIsGone` covers the aborted and replaced cases.

PR #685, opened 2026-08-17
fix Install Protoc

pfi79:


HagarMeir (commented):
This PR replaces the `arduino/setup-protoc` action with a manual `curl`/`unzip` install of protoc in `.github/workflows/build.yml`. CI-only change; the mechanics (version tag → asset name, bin/include layout) are correct. Findings are robustness/supply-chain/efficiency concerns:

**1. `curl` lacks `--fail`, so HTTP errors go undetected (build.yml:29)**
`curl -L` exits 0 even on a 404/5xx, writing the error HTML body into `/tmp/protoc.zip`. The `unzip` step then fails with a cryptic "cannot find zipfile directory" instead of a clear HTTP error. Most likely to bite on a future version bump. Fix: `curl -fL --retry 3`.

**2. No integrity check on the downloaded binary (build.yml:29)**
The protoc zip is fetched and executed to generate `.pb.go` with no sha256/signature verification — a supply-chain regression versus a pinned action. Fix: verify an expected sha256 before unzipping.

**3. Lost built-in caching (build.yml:27)**
`arduino/setup-protoc` cached protoc; the manual install re-downloads on every run, adding latency and a hard dependency on GitHub release availability/rate limits (unauthenticated). Consider `actions/cache` keyed on `PROTOC_VER`.

**4. Arch/OS hardcoded to `linux-x86_64` (build.yml:29)**
The action auto-detected platform; this hardcodes it. If `runs-on` later moves to arm64 or a matrix, it silently downloads the wrong-arch binary. Fix: derive arch/os from the runner.

**5. `include/` extracted but not on the search path (build.yml:31) — latent**
Only `$HOME/protoc/bin` is exposed; well-known-type resolution relies solely on protoc's relative `../include` lookup. Harmless today, but would break if a future `.proto` imports well-known types and the binary is ever relocated/symlinked. Optional: pass `-I "$HOME/protoc/include"`.

**Priorities:** #1 and #2 are worth addressing before merge (a one-line `-fL --retry` plus a checksum). #3–#5 are judgment calls to weigh against simplicity.

HagarMeir (approved):

